PR #30, opened 2018-12-20
Add method "SetMousePosition" to window class.

zlixine:
I noticed there's no function to change mouse position manually when I was trying to make ImGui use LLGL.

Notes:
-Works only for Windows.
-I wanted to implement it in Input, but I needed the window handle, some platform specific functions and also needed to call PostGlobalMotion, so I implemented it in window class instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .DS_Store
 CMakeFiles
 tests/Output
+/Projects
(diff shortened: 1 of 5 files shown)


LukasBanana:
Have you tried using the [`GetNativeHandle`](https://github.com/LukasBanana/LLGL/blob/master/include/LLGL/Surface.h#L48) function? Having a function in the `Window` interface that is restricted to Win32 platform is not really tempting.

zlixine:
It isn't really restricted, I implemented and tested it in Win32, as I don't have any active Linux or Mac machines.

LukasBanana:
Thank you for your contribution. Although such a function is generally handy, I currently don't see the necessity in that particular interface.
An interface should be designed to work with all platforms whenever possible and I won't be able to work on Windows platform for the next couple of months.
Hence I would like you to either use your custom branch or move this function into your engine directly.

PR #32, opened 2019-01-27
Remove "lib" prefix from dll names when building with mingw

wirx6:
This fixes `no renderer modules available on target platform` error when running examples built with mingw

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -552,6 +552,10 @@ elseif(UNIX)
     target_link_libraries(LLGL X11 pthread Xxf86vm Xrandr)
 endif()
 
+if(MINGW)
+    set_target_properties(LLGL PROPERTIES PREFIX "")
+endif()
+
 set_target_properties(LLGL PROPERTIES LINKER_LANGUAGE CXX DEBUG_POSTFIX "D")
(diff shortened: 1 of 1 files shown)


LukasBanana:
Thank you for your contribution.

I never considered using MinGW for this project, because D3D11 or later will be tough to compile with anything else but MSVC.
How do you compile these targets anyways?

I think in this case, however, it is better to simply construct the module name just like on the Linux platform (since MinGW is coming from the POSIX world), [see here](https://github.com/LukasBanana/LLGL/blob/master/sources/Platform/Linux/LinuxModule.cpp#L40).

This also ensures that MinGW compiled DLLs are not accidentally linked to an MSVC build.

If you agree, I would close this pull request and insert a small preprocessor compiler check à la
```c
#ifdef __MINGW32__
```

wirx6:
For now I'm testing OpenGL and Vulkan backends using MSYS2, D3D11 and D3D12 are not built.
I won't mind if you implement this fix differently.

LukasBanana:
Fixed with [d507920](https://github.com/LukasBanana/LLGL/commit/d507920d701a8b8efaa46d9217da146148690f0b).

PR #36, opened 2019-07-02
Fixed some win32 compile errors.

raxvan:
* added include <string> for std::wstring definition;
* changed win32 api to use 'W' version to fix compile errors when using 'Use Multi-Byte Character Set' option. When that option is used strings are expected to be char instead of wchar_t. see https://docs.microsoft.com/en-us/windows/desktop/LearnWin32/working-with-strings

### DIFF
--- a/include/LLGL/Display.h
+++ b/include/LLGL/Display.h
@@ -13,7 +13,7 @@
 #include "DisplayFlags.h"
 #include <vector>
 #include <memory>
-
+#include <string>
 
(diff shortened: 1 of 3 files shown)


LukasBanana:
Thanks for your contribution. I'm aware of the `A` and `W` versions of the Win32 API, but I apparently got used to the macros and never tried this 'Use Multi-Byte Character Set' option.

PR #39, opened 2019-07-07
Static compile and uniform empty api

raxvan:
* Static linking now supports multiple renderers in the same lib.
* Added empty api for shader uniforms.
Note:
Static linking will not not compile with current cmake setup. Each renderer api must now have one of the following defines:
- `LLGL_HAS_MODULE_DX11`
- `LLGL_HAS_MODULE_DX12`
- `LLGL_HAS_MODULE_OPENGL`
- `LLGL_HAS_MODULE_OPENGLES3`
- `LLGL_HAS_MODULE_VULKAN`
- `LLGL_HAS_MODULE_METAL`


LukasBanana:
Thanks for your contribution. It would be great if you could separate different changes into individual pull requests. It's quite a lot you have here at once. Can you do that please? For instance, one for static linking with multiple renderers etc.
I also worked on the shader uniforms today, so maybe you can have a look into the `push-constants` branch first.

Thanks
Lukas

raxvan:
> 
> 
> Thanks for your contribution. It would be great if you could separate different changes into individual pull requests. It's quite a lot you have here at once. Can you do that please? For instance, one for static linking with multiple renderers etc.
> I also worked on the shader uniforms today, so maybe you can have a look into the `push-constants` branch first.
> 
> Thanks
> Lukas

Yes, i should have, by the time i realized i had already pushed the changes on my fork.

LukasBanana:
I know this is not trivial with git, but there is definitely a way to separate this afterwards. At least you have multiple commits, so maybe you can create pull requests from these commits.

Please also have a look at the [project's coding guideline](https://github.com/LukasBanana/LLGL/blob/master/docu/CodingConventions/Coding%20Conventions%20for%20LLGL.pdf). At least tab characters should be avoided, I use 4 spaces instead. Tabs always look different depending on your IDE settings (or on github).

raxvan:
Updated the namespace names and converted all tabs to spaces (except stb image).

LukasBanana:
It would be good to start with reverting all the "refactoring" changes of blank lines between include headers to simplify this pull request. After doing all the changes that should not be part of the pull request (including the tabs), please squash those commits and then force push into your remote repository.

sources/Renderer/ModuleInterface.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+#include "ModuleInterface.h"
+#include <LLGL/RenderSystemFlags.h>
+
+#ifdef LLGL_BUILD_STATIC_LIB
+
+namespace LLGL
+{
+    
+#ifdef LLGL_HAS_MODULE_DX11
```
LukasBanana: Where are these macros defined? I can't find them in the `CMakeLists.txt` file.

include/LLGL/StreamOutputAttribute.h
```diff
@@ -11,6 +11,7 @@
 
 #include "Export.h"
 #include "Format.h"
+
```
LukasBanana: Generally not a problem for me when you do your formating here as you like, but in order to keep this pull request simpler, please revert all these blank lines between include headers you added.

sources/Renderer/ModuleInterface.cpp
raxvan: I use a custm build system. I'm not a familiar with cmake so i prefer to leave to someone else to do it instead of me breaking something there.

include/LLGL/StreamOutputAttribute.h
raxvan: They are intentional in this particular case, because i want to see what headers are included. Might change all includes in the future.

sources/Renderer/Metal/MTModuleInterface.mm
```diff
@@ -9,30 +9,55 @@
 #include "MTRenderSystem.h"
 
 
-extern "C"
+namespace LLGL
 {
+	namespace ModuleMetal
```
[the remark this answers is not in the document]
raxvan: I never had issues with mixed tabs and spaces, since i always use external tools. I ran a bash script to fix that , no idea why the tab was not fixed.

LukasBanana:
Would you please still revert your blank lines between the headers? Then squash your commits and we are good to go. If your merge from the master branch is making trouble with squasing your commits, you can also create a git patch and use that for a new pull request. Whatever is easier for you.

sources/Renderer/ModuleInterface.cpp
LukasBanana: Alright, I can take care of that.
Just out of curiosity: What kind of custom build system do you have?

raxvan: I use python to collect resources and i finally generate a premake5 "solution". Out of that i run premake5 to generate solution for xcode/vs/linux. It stuck with me because describing what and how to build tends to be a lot shorter and more flexible than cmake. For example llgl project only takes 130 lines of python code. Also there are some limitations with cmake that i previously ran into (compilation units, precompiled headers, magic errors, versions mismatch, etc). The code is not available publicly but you can see an example project here: https://github.com/raxvan/cllio/blob/master/prj.cllio.py

LukasBanana: Would be interesting to see alternative build systems for LLGL in the future, but it's probably more work to maintain multiple build systems at the same time. I once saw a [talk from Unity](https://www.youtube.com/watch?v=jmadc8xI_6I) about the build system Gradle. I never came to transition to this system though.

raxvan:
Is there a way to run Clang Format on all sources ?

raxvan:
I think it would be best to introduce .clang-format in the repository.

LukasBanana:
Looks like a reasonable tool, but I'm not a big fan of tools for automated code formatting. I tend to write my formatting style with some exceptions from time to time. It always annoys me when those tools discard my previous formatting.

LukasBanana:
After your merge with my master branch, rebasing has become difficult.
I could squash and force push your commits myself, but I'm not sure whether this complicates things even more for your repo. So I recommend you do it yourself and squash at least the last 4 commits. I'd like to avoid having multiple "removed unused code" commits after another. Here is what I did for testing on my local machine with a clone of your repo:
```
git rebase -i HEAD~4
```
Then edit the lower 3 commits as follows in order to squash them into the upmost commit:
```
pick 28b11c1 * removed unused code.
s 3c1645e * removed unused code.
s 6c0a7a7 * removed empty lines.
s 7de02e6 * removed empty lines.
```
Then force push your master branch to your remote repository.

For future pull requests, please remember to keep these pull requests simple, because some changes might be refected or need updates independently of other changes.

Thanks
Lukas

PR #41, opened 2019-07-14
* added support for multiple modules in static compile.

raxvan:


LukasBanana:
I think you forgot `ModuleInterface.cpp` and the updated `ModuleInterface.h` files.
Just add them to fix the failed CI builds and I merge it.

Regarding your request for "clang format": the only script for a few settings I have is `.editorconfig`, but I guess it's only supported by VS2017 or later.

raxvan:
Added missing file and merge with latest changes master.

PR #50, opened 2020-01-15
add include directory to cmake LLGL target

underdoeg:
This allows to use LLGL as a submodule and library.

For example:
```
add_executable(test main.cpp)
target_link_libraries(projection LLGL)
```

Now #include <LLGL/LLGL.h> works from test.cpp

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -600,6 +600,8 @@ else()
     add_library(LLGL SHARED ${FilesLLGL})
 endif()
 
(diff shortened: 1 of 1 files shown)


LukasBanana:
Looks good, but wouldn't it make sense to even replace the `include_directories("${PROJECT_INCLUDE_DIR}")` entirely by this project specific include path?

underdoeg:
Yes, I think it would. I just submitted the PR with the most non intrusive fix I could find.

PR #51, opened 2020-01-18
Use native FindVulkan.cmake

julianxhokaxhiu:
**Backstory:**
I was trying to build this library under Visual Studio 2019 for Win32 Arch and I found out that Vulkan was not able to be found, with the latest SDK version ( 1.1.130 ). Looking at how you did include Vulkan I found out you're using the old suggested method. Nowadays it's better to use the native `FindVulkan.cmake` which is provided by any Cmake version that is >= 3.7.

I doubt someone will build this code with an older CMake version.

Thank you for the awesome work you're doing. Looking forward on this project.


julianxhokaxhiu:
Sorry for the subsequent commits but it was unplanned as I thought it was all about fixing the `CMakeLists.txt` file, while it ended being a more sophisticated patch.

Overall I managed to use the Vulkan SDK on LInux Travis CI jobs, instead of Mesa Vulkan library ( allowing this library to be truly Vulkan cross-platform ). Although for some reasons the Linux job still fails in building the Vulkan code, while on Windows I do not have any kind of issues building the same code under VS2019 using CMake.

Feel free to squash this PR when merging.

.travis.yml
```diff
@@ -8,9 +8,12 @@ matrix:
...
       before_install:
       - git clone https://github.com/LukasBanana/GaussianLib.git
-      - sudo apt-get install libx11-dev libxxf86vm-dev libxrandr-dev libglu1-mesa-dev freeglut3-dev mesa-common-dev libvulkan1 mesa-vulkan-drivers vulkan-utils
+      - wget -qO - http://packages.lunarg.com/lunarg-signing-key-pub.asc | sudo apt-key add -
+      - sudo wget -qO /etc/apt/sources.list.d/lunarg-vulkan-1.1.130-xenial.list http://packages.lunarg.com/vulkan/1.1.130/lunarg-vulkan-1.1.130-xenial.list
+      - sudo apt update
+      - sudo apt-get install libx11-dev libxxf86vm-dev libxrandr-dev libglu1-mesa-dev freeglut3-dev mesa-common-dev vulkan-sdk vulkan-tools
       before_script:
-      - cmake -DDLLGL_BUILD_RENDERER_OPENGL=ON -DLLGL_BUILD_RENDERER_VULKAN=ON -DLLGL_BUILD_EXAMPLES=ON -DGaussLib_INCLUDE_DIR:STRING="GaussianLib/include" -S .
+      - cmake -DLLGL_BUILD_RENDERER_OPENGL=ON -DLLGL_BUILD_RENDERER_VULKAN=ON -DLLGL_BUILD_EXAMPLES=ON -DGaussLib_INCLUDE_DIR:STRING="GaussianLib/include" -S .
```
LukasBanana: Good catch! IIRC this option is enabled by default and that's why I haven't seen this typo. But I wanted to make it explicit which renderer is enabled, so that's fine.

CMakeLists.txt
```diff
@@ -674,9 +674,9 @@ endif()
 
 if(NOT APPLE AND LLGL_BUILD_RENDERER_VULKAN)
     # Vulkan Renderer
-    include(cmake/FindVulkan.cmake)
-    if(VULKAN_FOUND)
-        include_directories(${VULKAN_INCLUDE_DIR})
+    find_package(Vulkan REQUIRED)
```
LukasBanana: Using the native `find_package` modules is definitely the preferred way to provide better cross-platform support 👍

julianxhokaxhiu:
@LukasBanana Done! Let me know if you want also the Release patch into this PR or if you're doing to fix it yourself.

LukasBanana:
Thank you for your contribution.

Regarding your Release patch: why do you need `<stdexcept>` in `VKDeviceBuffer.h`?

julianxhokaxhiu:
@LukasBanana this is why: https://ci.appveyor.com/project/julianxhokaxhiu/llglci/build/job/fku2on40dcxatftt#L300

LukasBanana:
Well then feel free to make another PR for that fix. Thank you

PR #57, opened 2020-07-03
Fixed large memory leaks in the Metal sample programs

beldenfox:
The sample programs were leaking memory on every frame because autorelease pools were not being created or drained so Metal objects were not being released. I added a DrawNow call to the RenderContext to give the platform context an opportunity to set up the correct environment for drawing and clean up after the drawing is done.

I also added an autorelease pool inside OnProcessEvents and fixed a bug in MTEncodeScheduler that was leaking copied
RenderPassDescriptors.

### DIFF
--- a/examples/Cpp/ExampleBase/ExampleBase.cpp
+++ b/examples/Cpp/ExampleBase/ExampleBase.cpp
@@ -230,7 +230,14 @@ void ExampleBase::Run()
         }
 
         // Draw current frame
+        #ifdef LLGL_OS_MACOS
+        @autoreleasepool
+        {
+            OnDrawFrame();
+        }
+        #else
         OnDrawFrame();
+        #endif
 
         // Check if resolution has changed
         auto currentResolution = context->GetResolution();
--- a/sources/Platform/MacOS/MacOSWindow.mm
+++ b/sources/Platform/MacOS/MacOSWindow.mm
@@ -496,23 +496,26 @@ NSWindow* MacOSWindow::CreateNSWindow(const WindowDescriptor& desc)
 
 void MacOSWindow::OnProcessEvents()
 {
-    NSEvent* event = nil;
-
-    /* Process NSWindow events with latest event types */
-    while ((event = [wnd_ nextEventMatchingMask:g_EventMaskAny untilDate:nil inMode:NSDefaultRunLoopMode dequeue:YES]) != nil)
-        ProcessEvent(event);
-
-    /* Check for window signales */
-    if ([(MacOSWindowDelegate*)[wnd_ delegate] popResizeSignal])
+    @autoreleasepool
     {
-        /* Get size of the NSWindow's content view */
-        NSRect frame = [[wnd_ contentView] frame];
+        NSEvent* event = nil;
 
-        auto w = static_cast<std::uint32_t>(frame.size.width);
-        auto h = static_cast<std::uint32_t>(frame.size.height);
+        /* Process NSWindow events with latest event types */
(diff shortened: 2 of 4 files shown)


sources/Renderer/Metal/MTEncoderScheduler.mm
```diff
@@ -101,6 +101,7 @@
             renderPassDesc.stencilAttachment.loadAction = MTLLoadActionLoad;
         }
         BindRenderEncoder(renderPassDesc);
+        [renderPassDesc release];
```
LukasBanana, line 104: That one here looks ok to me. I'm sometimes not sure when to use `release` and in Obj-C.

examples/Cpp/ExampleBase/ExampleBase.cpp
```diff
@@ -230,7 +230,7 @@ void ExampleBase::Run()
         }
 
         // Draw current frame
-        OnDrawFrame();
+        context->DrawNow([=]{OnDrawFrame();});
```
LukasBanana, line 233: It appears to me we can fix this by simple wrapping this call in an `@autoreleasepool` like this:
```cpp
#ifdef LLGL_PLATFORM_MACOS
@autoreleasepool {
    OnDrawFrame();
}
#else
OnDrawFrame();
#endif
```
It would of course need to be in a separate file such as `ExampleBaseFrame.mm` to provide Obj-C compatibility.

beldenfox: You're right, I'm over-engineering this based on past experience. In similar projects I've always run into issues when drawing outside of an OS-issued update event. There's always some bit of bookkeeping to be done before and after the drawing operation to make this work reliably e.g. correctly managing HDC's on WIN32 or autorelease pools on Mac. I've also needed a call like DrawNow in the past when working with multiple drawing contexts in the same window (something I'm looking into with LLGL). But for now it's overkill for the somewhat artificial use case of these sample programs.

A more localized fix is to tweak the .cpp file and then alter the CMakeLists.txt file to force it to be compiled as Objective-c++. Then you won't need a separate .mm file for the Mac. I'll look into that.

LukasBanana: > A more localized fix is to tweak the .cpp file and then alter the CMakeLists.txt file to force it to be compiled as Objective-c++
That's also an option, but I'm trying to keep the CMake files simple. I worked with projects where the CMake script executes Python scripts and other indirections.

LukasBanana:
Thanks for the update. I can test this on my Mac tomorrow.

examples/Cpp/ExampleBase/ExampleBase.cpp
```diff
@@ -230,7 +230,13 @@ void ExampleBase::Run()
         }
 
         // Draw current frame
+        #ifdef LLGL_OS_MACOS
+        @autoreleasepool {
```
LukasBanana, line 234: Please apply the braces according to the [Allman coding style](https://en.wikipedia.org/wiki/Indentation_style#Allman_style) also for the `autoreleasepool` statement. Rest of the PR looks good.

sources/Platform/MacOS/MacOSWindow.mm
```diff
@@ -496,23 +496,25 @@ static NSUInteger GetNSWindowStyleMask(const WindowDescriptor& desc)
 
 void MacOSWindow::OnProcessEvents()
 {
-    NSEvent* event = nil;
+    @autoreleasepool {
```
LukasBanana, line 499: Same here as above. Please adjust the braces to the Allman coding style, sqaush your commits and I'll merge the PR.

PR #59, opened 2020-07-10
Fixed-index primitive restart should now work in OpenGL version 3.1 and up.

beldenfox:
Added a sample program which exercises the fixed-index primitive restart feature. On OpenGL before 4.3 this has to be implemented by enabling GL_PRIMITIVE_RESTART and then calling glPrimitiveRestartIndex to ensure the index is set appropriately for the width of the current index buffer entries.

This pull request is primarily a design review. The question is when to update the restart index to match the width of the index buffer entries. My approach was to do this at the point where the index buffer becomes "active" i.e. bound to the currently bound VAO (in whatever order those bindings occur). I think I tracked down the point in the code where this happens but am not entirely sure.

An alternative was to track down every gl*Draw*Elements* call and update the restart index just before the call. This would ensure that the restart index was only updated when absolutely necessary but that approach involved more changes and didn't seem as maintainable.


examples/Cpp/PrimitiveRestart/Example.140core.frag
```diff
@@ -0,0 +1,14 @@
+// GLSL shader version 1.40 (for OpenGL 3.2 Core Profile)
```
LukasBanana, line 1: Typo: GLSL 140 was introduced in OpenGL 3.1 Core Profile.

examples/Cpp/PrimitiveRestart/Example.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+    // Vertex data structure
+    struct Vertex
+    {
+        Gs::Vector2f        position;
+        LLGL::ColorRGBAub   color;
+    };
+
+    template <typename T>
+    static void addSquare(float centerX, float centerY, float size, std::vector<Vertex>& vertices, std::vector<T>& indices)
```
LukasBanana, line 42: Please don't use _lower camel case_ for functions in this project. I wrote a coding style guideline for this project a while back: [LLGL Coding Conventions](https://github.com/LukasBanana/LLGL/blob/master/docu/CodingConventions/Coding%20Conventions%20for%20LLGL.pdf). The rest looks very good 👍

examples/Cpp/PrimitiveRestart/Example.cpp
```diff
@@ -0,0 +1,177 @@
...
+        float bottom = centerY - size / 2.0;
+
+        size_t startIndex = vertices.size();
+        vertices.push_back({{ right, top    }, { 255,   0,   0, 255 } });
+        vertices.push_back({{ right, bottom }, {   0, 255,   0, 255 } });
+        vertices.push_back({{ left,  top    }, {   0,   0, 255, 255 } });
+        vertices.push_back({{ left,  bottom }, { 255, 255, 255, 255 } });
+
+        if (!indices.empty())
+            indices.push_back((T) 0xFFFFFFFF);
```
LukasBanana, line 56: Please use the constructor expression here, i.e. `T(0xFFFFFFFF)`. I'm not a fan of C-style casts as they are error prone and nasty to search for. I don't mind if they are in any of the third party dependencies though. In this case it doesn't really matter, but I'd like to avoid them completely.

examples/Cpp/PrimitiveRestart/Example.cpp
```diff
@@ -0,0 +1,177 @@
...
+    {
+        // Vertex data, an array of 16 points that we'll use
+        // to draw triangles.
+        const float s = 0.5f;
+
+        std::vector<Vertex> vertices;
+        std::vector<uint16_t> indices16;
+        std::vector<uint32_t> indices32;
+
+        addSquare( .5,  .5, 0.8, vertices, indices16);
```
LukasBanana, line 73: [MSVC emits warnings here](https://ci.appveyor.com/project/LukasBanana/llgl/builds/34031207/job/q1mfig2h0l6a5umd#L191) because of `double` literals used for `float` parameters in a template. Simply adding the `f` suffix should fix that.

examples/Cpp/PrimitiveRestart/Example.cpp
```diff
@@ -0,0 +1,177 @@
...
+        indices.push_back(startIndex + 1);
+        indices.push_back(startIndex + 2);
+        indices.push_back(startIndex + 3);
+    }
+
+    std::vector<LLGL::VertexFormat> CreateBuffers()
+    {
+        // Vertex data, an array of 16 points that we'll use
+        // to draw triangles.
+        const float s = 0.5f;
```
LukasBanana, line 67: It seems this constant is not in use, which [Codacy is complaining about](https://app.codacy.com/manual/LukasBanana/LLGL/pullRequest?prid=5835647).

examples/Cpp/PrimitiveRestart/Example.cpp
```diff
@@ -0,0 +1,177 @@
...
+ * Example.cpp (Example_PrimitiveRestart)
+ *
+ * This file is part of the "LLGL" project (Copyright (c) 2015-2019 by Lukas Hermanns)
+ * See "LICENSE.txt" for license information.
+ */
+
+#include <ExampleBase.h>
+#include <vector>
+
+class Example_PrimitiveRestart : public ExampleBase
```
LukasBanana, line 11: Feel free to add your name to this example in a comment like `// author: ...` if you want to :)

sources/Renderer/OpenGL/GLCoreProfile/GLCoreProfileCaps.cpp
```diff
@@ -224,12 +224,16 @@ static void GLGetFeatureLimits(RenderingLimits& limits)
...
     limits.maxPatchVertices                 = GLGetUInt(GL_MAX_PATCH_VERTICES);
     limits.maxAnisotropy                    = static_cast<std::uint32_t>(GLGetFloat(GL_MAX_TEXTURE_MAX_ANISOTROPY_EXT));
-    limits.maxComputeShaderWorkGroups[0]    = GLGetUIntIndexed(GL_MAX_COMPUTE_WORK_GROUP_COUNT, 0);
-    limits.maxComputeShaderWorkGroups[1]    = GLGetUIntIndexed(GL_MAX_COMPUTE_WORK_GROUP_COUNT, 1);
-    limits.maxComputeShaderWorkGroups[2]    = GLGetUIntIndexed(GL_MAX_COMPUTE_WORK_GROUP_COUNT, 2);
-    limits.maxComputeShaderWorkGroupSize[0] = GLGetUIntIndexed(GL_MAX_COMPUTE_WORK_GROUP_SIZE, 0);
-    limits.maxComputeShaderWorkGroupSize[1] = GLGetUIntIndexed(GL_MAX_COMPUTE_WORK_GROUP_SIZE, 1);
-    limits.maxComputeShaderWorkGroupSize[2] = GLGetUIntIndexed(GL_MAX_COMPUTE_WORK_GROUP_SIZE, 2);
+
+    if (features.hasComputeShaders)
```
LukasBanana, line 228: Good catch 👍

examples/Cpp/PrimitiveRestart/Example.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+    template <typename T>
+    static void addSquare(float centerX, float centerY, float size, std::vector<Vertex>& vertices, std::vector<T>& indices)
+    {
+        float left = centerX - size / 2.0;
+        float right = centerX + size / 2.0;
+        float top = centerY + size / 2.0;
+        float bottom = centerY - size / 2.0;
+
+        size_t startIndex = vertices.size();
```
LukasBanana, line 49: `startIndex` should be of type `T` because [MSVC 2019 Win64 is complaining](https://ci.appveyor.com/project/LukasBanana/llgl/builds/34031207/job/50f2t2pm0u3kkte6#L204) about implicit truncation from `size_t` to `T`. I suggest something like this:
```cpp
const T startIndex = static_cast<T>(vertices.size());
```
The conversion between `size_t` and `uint32_t` can be a bit of a pain in LLGL, but 32 bit unsigned integers are all we get (most of the time) from rendering APIs up until Vulkan and D3D12.

sources/Renderer/OpenGL/RenderState/GLStateManager.cpp
```diff
@@ -550,6 +550,18 @@ void GLStateManager::SetLineWidth(GLfloat width)
...
 }
 
+void GLStateManager::SetPrimitiveRestartIndex(bool indexType16Bits)
+{
+    #ifdef LLGL_PRIMITIVE_RESTART
+    GLuint index = (indexType16Bits ? 0xFFFF : 0xFFFFFFFF);
+    if (commonState_.primitiveRestartIndex != index)
+    {
+        commonState_.primitiveRestartIndex = index;
+        glPrimitiveRestartIndex(index);
```
LukasBanana, line 560: This function looks good, but we'll have to add `glPrimitiveRestartIndex` to the declaration of GL extensions (see [GLCoreExtensionsDecl.inc](https://github.com/LukasBanana/LLGL/blob/master/sources/Renderer/OpenGL/GLCoreProfile/GLCoreExtensionsDecl.inl)). This is not necessary for MacOS, but for the other platforms. I'm sometimes having a hard time finding the right extension either on the [OpenGL registry page](https://www.khronos.org/registry/OpenGL/index_gl.php) or via a search engine. If you can't find it either, maybe you can add another section for `GL_VERSION_3_1` like the one for [`glGetStringi`](https://github.com/LukasBanana/LLGL/blob/master/sources/Renderer/OpenGL/GLCoreProfile/GLCoreExtensionsDecl.inl#L66).

If you need help with that part, I can take care of it after I merged the PR.

examples/Cpp/PrimitiveRestart/Example.140core.frag
beldenfox: The original mismatch is in the HelloTriangle sample. I just grabbed those shaders and bumped both the OpenGL and GLSL versions.

sources/Renderer/OpenGL/RenderState/GLStateManager.cpp
beldenfox: From what I can tell primitive restart went straight from an NV extension (GL_NV_PRIMITIVE_RESTART) to 3.1 core. I took a shot at updating this file but don't have a good way of testing my changes, I only code on Mac and Raspberry Pi these days.

beldenfox:
I cleaned up the screenshot. I now realize that an 800x600 window will produce a 1600x1200 screenshot on a Retina display (plus more for the title bar and such) which makes sense given how Retina works.

examples/Cpp/PrimitiveRestart/Example.140core.frag
LukasBanana: Thanks, I'll fix those files myself then.

LukasBanana:
@beldenfox I submitted a few updates with [8065922](https://github.com/LukasBanana/LLGL/commit/8065922b2cf762facffd73eec2433bedcf4c2c77) especially to address the loading of `glPrimitiveRestartIndex` which seems to be part of the `GL_ARB_compatibility` extension (which is why it's hard to find it online).

Your example also reveals that the primitive restart index for 16 bit index buffers doesn't work for D3D12 (see attached screenshot), but that's on me. Unfortunately though, I currently don't have an elegant solution for that.

21_PrimitiveRestart_ErrorOnD3D12

PR #60, opened 2020-07-21
Correctly handling Retina displays on the Mac

beldenfox:
I’ve been working on getting Retina support into LLGL. The new code is not done and requires a lot of clean-up but the changes involve re-thinking what SetVideoMode does so I thought you should take a look sooner rather than later.

In Retina mode the size of the surface and the size of the drawables diverge. An NSWindow is sized in a space that roughly approximates 72 dpi and then the drawables are sized upward to a match the actual dpi of the screen. So I needed to alter SetVideoMode to resize the drawables independently of the surface size. I started by stubbing out Window::AdjustForVideoMode and all of the fullscreen handling (more on fullscreen later).

Then I added Surface::GetPreferredResolution() so a surface can advertise a preferred resolution that differs from GetContentSize(). Finally I updated OnSetVideoMode for both Metal and OpenGL to resize the drawables appropriately. And that was pretty much it beyond tweaking Info.plist and updating the sample code a bit.

Basically I’ve pared down SetVideoMode to only update the size of the drawables, roughly equivalent to ResizeBuffer in DirectX. This eliminates the possibility that calling SetVideoMode could generate a resize event while handling a resize event. It also eliminates the code that was trying to re-center the window on every resize event which was making live re-resizing impossible on the Mac.

(BTW, I tweaked the DirectX versions of OnSetVideoMode to allow arbitrary sizes to match the Mac. On Windows the main use would be to to reduce the size of the drawables for performance reasons.)

The next step is to get fullscreen working again. Previously SetVideoMode was called at both ends of the process, first to kick the system into fullscreen mode and then to handle the resulting resize event. I want to avoid that recursion by introducing new RenderContext routines for entering and exiting fullscreen mode. Then clean-up, including adding some capability flags.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,6 +44,8 @@ endmacro()
 macro(ADD_EXAMPLE_PROJECT TEST_NAME TEST_FILES LIB_FILES)
     if(APPLE)
         add_executable(${TEST_NAME} MACOSX_BUNDLE ${TEST_FILES})
+        set_target_properties(${TEST_NAME} PROPERTIES
+            MACOSX_BUNDLE_INFO_PLIST ${EXAMPLE_PROJECTS_DIR}/MacOS/HighResolution.plist)
     elseif(LLGL_ANDROID_PLATFORM)
         add_library(${TEST_NAME} SHARED ${TEST_FILES})
         target_link_libraries(${TEST_NAME} log)
--- a/examples/Cpp/ExampleBase/ExampleBase.h
+++ b/examples/Cpp/ExampleBase/ExampleBase.h
@@ -170,7 +170,7 @@ protected:
 
     ExampleBase(
         const std::wstring&     title,
-        const LLGL::Extent2D&   resolution  = { 800, 600 },
+        const LLGL::Extent2D&   contentSize = { 800, 600 },
         std::uint32_t           samples     = 8,
         bool                    vsync       = true,
         bool                    debugger    = true
--- a/examples/Cpp/MacOS/HighResolution.plist
+++ b/examples/Cpp/MacOS/HighResolution.plist
@@ -1,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>NSHighResolutionCapable</key>
+	<true/>
+	<key>CFBundleDevelopmentRegion</key>
+	<string>English</string>
+	<key>CFBundleExecutable</key>
+	<string>${MACOSX_BUNDLE_EXECUTABLE_NAME}</string>
+	<key>CFBundleGetInfoString</key>
+	<string>${MACOSX_BUNDLE_INFO_STRING}</string>
+	<key>CFBundleIconFile</key>
+	<string>${MACOSX_BUNDLE_ICON_FILE}</string>
+	<key>CFBundleIdentifier</key>
+	<string>${MACOSX_BUNDLE_GUI_IDENTIFIER}</string>
+	<key>CFBundleInfoDictionaryVersion</key>
+	<string>6.0</string>
+	<key>CFBundleLongVersionString</key>
+	<string>${MACOSX_BUNDLE_LONG_VERSION_STRING}</string>
+	<key>CFBundleName</key>
+	<string>${MACOSX_BUNDLE_BUNDLE_NAME}</string>
+	<key>CFBundlePackageType</key>
+	<string>APPL</string>
+	<key>CFBundleShortVersionString</key>
+	<string>${MACOSX_BUNDLE_SHORT_VERSION_STRING}</string>
+	<key>CFBundleSignature</key>
+	<string>????</string>
+	<key>CFBundleVersion</key>
+	<string>${MACOSX_BUNDLE_BUNDLE_VERSION}</string>
+	<key>CSResourcesFileMapped</key>
+	<true/>
+	<key>NSHumanReadableCopyright</key>
+	<string>${MACOSX_BUNDLE_COPYRIGHT}</string>
+</dict>
+</plist>
--- a/sources/Platform/MacOS/MacOSWindow.mm
+++ b/sources/Platform/MacOS/MacOSWindow.mm
@@ -143,8 +143,6 @@ static const auto g_EventTypeScrollWheel        = NSScrollWheel;
 
 - (void)windowDidResize:(NSNotification*)notification
 {
-    //TODO: callback (here PostResize) must currently not be called while the NSEvent polling has not finished!
-    #if 0
     /* Get size of the NSWindow's content view */
     NSWindow* sender = [notification object];
     NSRect frame = [[sender contentView] frame];
@@ -154,9 +152,6 @@ static const auto g_EventTypeScrollWheel        = NSScrollWheel;
 
     /* Notify event listeners about resize */
     window_->PostResize({ w, h });
-    #else
-    resizeSignaled_ = YES;
-    #endif
 }
 
 - (NSApplicationPresentationOptions)window:(NSWindow*)window willUseFullScreenPresentationOptions:(NSApplicationPresentationOptions)proposedOptions
@@ -284,6 +279,28 @@ Extent2D MacOSWindow::GetContentSize() const
     return GetSize(true);
 }
 
+Extent2D MacOSWindow::GetPixelResolution() const
+{
+    auto size = wnd_.contentView.bounds.size;
+    size.width *= wnd_.backingScaleFactor;
+    size.height *= wnd_.backingScaleFactor;
+
+    return Extent2D
+    {
+        static_cast<std::uint32_t>(size.width + 0.5),
+        static_cast<std::uint32_t>(size.height + 0.5)
+    };
+}
+
+Extent2D MacOSWindow::GetContentSizeForPixelResolution(const Extent2D& resolution) const
+{
+    return Extent2D
+    {
+        static_cast<std::uint32_t>(resolution.width / wnd_.backingScaleFactor + 0.5),
+        static_cast<std::uint32_t>(resolution.height / wnd_.backingScaleFactor + 0.5)
+    };
+}
+
 void MacOSWindow::SetPosition(const Offset2D& position)
 {
     /* Get visible screen size (without dock and menu bar) */
@@ -394,13 +411,10 @@ void MacOSWindow::SetDesc(const WindowDescriptor& desc)
             [wnd_ setCollectionBehavior:NSWindowCollectionBehaviorDefault];
         #endif
 
-        //TOOD: incomplete -> must be ignored right now, otherwise window is moved on a resize event
-        #if 0
         if (desc.centered)
             [wnd_ center];
         else
             SetPosition(desc.position);
-        #endif
 
         SetSize(desc.size);
         SetTitle(desc.title);
--- a/sources/Renderer/Direct3D11/D3D11RenderContext.cpp
+++ b/sources/Renderer/Direct3D11/D3D11RenderContext.cpp
@@ -110,6 +110,15 @@ bool D3D11RenderContext::OnSetVideoMode(const VideoModeDescriptor& videoModeDesc
     return true;
 }
 
+bool D3D11RenderContext::OnSetDrawableResolution(const Extent2D& resolution)
+{
+    //TODO It would be sufficient to resize the existing resources.
+    auto newMode = GetVideoMode();
+    newMode.resolution = resolution;
+    ResizeBackBuffer(newMode);
+    return true;
+}
+
 bool D3D11RenderContext::OnSetVsync(const VsyncDescriptor& vsyncDesc)
 {
     swapChainInterval_ = (vsyncDesc.enabled ? std::max(1u, std::min(vsyncDesc.interval, 4u)) : 0u);
--- a/sources/Renderer/Direct3D12/D3D12RenderContext.cpp
+++ b/sources/Renderer/Direct3D12/D3D12RenderContext.cpp
@@ -192,6 +192,16 @@ bool D3D12RenderContext::OnSetVideoMode(const VideoModeDescriptor& videoModeDesc
     return true;
 }
 
+bool D3D12RenderContext::OnSetDrawableResolution(const Extent2D& resolution)
+{
+    //TODO It would be sufficient to resize the existing resources.
+    auto newMode = GetVideoMode();
+    newMode.resolution = resolution;
+    /* Re-create resource that depend on the window size */
+    CreateWindowSizeDependentResources(newMode);
+    return true;
+}
+
 bool D3D12RenderContext::OnSetVsync(const VsyncDescriptor& vsyncDesc)
 {
     swapChainInterval_ = (vsyncDesc.enabled ? std::max(1u, std::min(vsyncDesc.interval, 4u)) : 0u);
--- a/sources/Renderer/Metal/MTRenderContext.mm
+++ b/sources/Renderer/Metal/MTRenderContext.mm
@@ -62,12 +62,18 @@ MTRenderContext::MTRenderContext(
     /* Initialize color and depth buffer */
     //MTLPixelFormat colorFmt = metalLayer_.pixelFormat;
 
+    // We have no need of the internal drawing timer
+    view_.paused                    = YES;
+    view_.autoResizeDrawable        = NO;
+
     view_.colorPixelFormat          = renderPass_.GetColorAttachments()[0].pixelFormat;
     view_.depthStencilPixelFormat   = renderPass_.GetDepthStencilFormat();
     view_.sampleCount               = renderPass_.GetSampleCount();
 
     if (desc.vsync.enabled)
         view_.preferredFramesPerSecond = static_cast<NSInteger>(desc.vsync.refreshRate);
+
+    OnSetDrawableResolution(desc.videoMode.resolution);
 }
 
 void MTRenderContext::Present()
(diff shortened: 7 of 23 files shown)


LukasBanana:
So far I was only manually adding the HighResolution attribute to the `.plist` files when testing something on Mac locally, but that was of course only a temporary solution. You seem to be more experienced with the Retina display support on the Mac. That being said, some of your changes seem to be quite temporary, too, like commenting out members from interface structures such as `fullscreen`.
I can take a closer look over the weekend, but if I understand you correctly, you're not quite done yet anyway.
Nevertheless, I'm happy to see someone working on those missing features in that project.

examples/Cpp/MacOS/HighResolution.plist
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
LukasBanana, line 1: See comment above: is this `.plist`file backwards compatible for older Macs without a retina display?

sources/Renderer/Direct3D12/D3D12RenderContext.cpp
```diff
@@ -185,9 +185,12 @@ bool D3D12RenderContext::OnSetVideoMode(const VideoModeDescriptor& videoModeDesc
     if (prevVideoMode.resolution != videoModeDesc.resolution)
         CreateWindowSizeDependentResources(videoModeDesc);
 
+#if 0
```
LukasBanana, line 188: See comment above: please let's focus on Mac, not on D3D in this PR.

sources/Renderer/Direct3D12/D3D12RenderContext.cpp
```diff
@@ -267,7 +270,7 @@ void D3D12RenderContext::CreateWindowSizeDependentResources(const VideoModeDescr
             swapChainDesc.SampleDesc.Quality    = 0;
             swapChainDesc.BufferUsage           = DXGI_USAGE_RENDER_TARGET_OUTPUT;
             swapChainDesc.BufferCount           = numFrames_;
-            swapChainDesc.Scaling               = DXGI_SCALING_NONE;
+            swapChainDesc.Scaling               = DXGI_SCALING_STRETCH;
```
LukasBanana, line 273: What's the benefit of this scaling option?

sources/Renderer/Metal/MTRenderContext.mm
```diff
@@ -62,12 +62,18 @@
     /* Initialize color and depth buffer */
     //MTLPixelFormat colorFmt = metalLayer_.pixelFormat;
 
+    // We have no need of the internal drawing timer
```
LukasBanana, line 65: That's an interesting change, too. Can you explain briefly what these two options do? Do we need to 'unpause' the view later?

sources/Renderer/Metal/MTRenderContext.mm
```diff
@@ -102,7 +108,25 @@
 
 bool MTRenderContext::OnSetVideoMode(const VideoModeDescriptor& videoModeDesc)
 {
-    return true; // do nothing
+    // There is only one scale factor. Verify that it works on both axes.
+    CGFloat scale = videoModeDesc.resolution.width / view_.bounds.size.width;
```
LukasBanana, line 112: I'm not experienced enough with the MacOS view/view-controller or `CAMetalLayer` system (or whatever type `view_.layer` is) to provide a sophisticated review on this function, but it seems to do its job, so looks good to me. The only request I have for this one is that I prefer to be more explicit about type casts:
```cpp
uint32_t scaledHeight = static_cast<uint32_t>(view_.bounds.size.height * scale + 0.5);
```

sources/Renderer/OpenGL/GLRenderContext.cpp
```diff
@@ -108,10 +108,16 @@ bool GLRenderContext::OnSetVideoMode(const VideoModeDescriptor& videoModeDesc)
     /* Notify GL context of a resize */
     context_->Resize(videoModeDesc.resolution);
 
+    // XXX TODO Changing resolution by altering the display configuration
```
LukasBanana, line 111: Again: please don't disable working features in other backends.

sources/Renderer/OpenGL/Platform/MacOS/MacOSGLContext.mm
```diff
@@ -68,6 +70,15 @@
 
 void MacOSGLContext::Resize(const Extent2D& resolution)
 {
+    GLint dim[2];
```
LukasBanana, line 73: Looks good to me, but again, I'm not experienced enough with the CGL functions to provide a sophisticated feedback on this one. But I'd appreciate a bit explanation.

sources/Renderer/RenderContext.cpp
```diff
@@ -86,44 +86,13 @@ bool RenderContext::SetVsync(const VsyncDescriptor& vsyncDesc)
...
-            windowDesc.centered         = !videoModeDesc.fullscreen;
-            windowDesc.windowContext    = windowContext;
-        }
-        surface_ = Window::Create(windowDesc);
-
-        #endif
-    }
-
-    /* Store video mode settings */
+    // XXX TODO
```
LukasBanana, line 89: I would like to find another solution that still provides a default surface. Are you referring to the information for resolution scaling or the preferred resolution? I think that should be doable.

sources/Renderer/RenderContext.cpp
```diff
@@ -133,6 +102,9 @@ void RenderContext::ShareSurfaceAndConfig(RenderContext& other)
     vsyncDesc_      = other.vsyncDesc_;
 }
 
+# if 0
+// XXX TODO On Mac and Windows platforms altering the user's display configuration
```
LukasBanana, line 106: Altering the user's display configuration works fine on Windows. Even when the application crashes, Windows resets the previous configuration. Only Linux has problems there when an application terminates abnormally. But for any rendering API on Windows other than D3D, you don't have much of a choice. Only D3D provides that functionality via the graphics API as far as I know.

sources/Renderer/RenderContext.cpp
```diff
@@ -180,9 +152,15 @@ bool RenderContext::SetVideoModePrimary(const VideoModeDescriptor& videoModeDesc
 
     auto& surface = GetSurface();
 
+    // XXX TODO We should not assume that a surface can enter fullscreen or even
```
LukasBanana, line 155: The manual store and restore of the surface location is indeed not a very elegant solution. But again, please don't just disable those code paths and please try to keep the PR small.

CMakeLists.txt
```diff
@@ -44,6 +44,17 @@ endmacro()
 macro(ADD_EXAMPLE_PROJECT TEST_NAME TEST_FILES LIB_FILES)
     if(APPLE)
         add_executable(${TEST_NAME} MACOSX_BUNDLE ${TEST_FILES})
+        set_target_properties(${TEST_NAME} PROPERTIES
+            MACOSX_BUNDLE_INFO_PLIST ${EXAMPLE_PROJECTS_DIR}/MacOS/HighResolution.plist)
+        # When debugging under Xcode we want changes to the renderer sources
```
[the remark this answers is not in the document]
beldenfox: Got it.

sources/Renderer/RenderContext.cpp
beldenfox: Altering the user's display configuration is frowned upon on the Mac and Apple does not believe it's necessary for performance reasons. And on Windows more apps are supporting borderless windowed fullscreen mode. In any case, the issue is that renderer back-ends can't opt out of this behavior. Even the D3D renderer's are not allowed to manipulate the display resolution using the D3D API calls.

sources/Renderer/RenderContext.cpp
beldenfox: By default surface you mean a Window which strikes me as rather focused on games which always fill the full window (shades of SDL). I would like to generalize LLGL so it can be used in other context e.g. multiple NSViews, individual CALayer's. In the app I'm working on I don't even have a need for the Window surface at all. But I'll think about it.

sources/Renderer/Metal/MTRenderContext.mm
beldenfox: According to the Apple documentation MTKViews by default run an internal timer trying to redraw the view on a fixed interval. Setting 'paused' to true just signals to the view that we don't want that timer. This is primarily to configure the view based on Apple's guidelines and let it know we're drawing whenever we want to.
The other option disables the view's default management of the drawable sizes. This allows us to directly set the resolution to anything we want.

sources/Renderer/Direct3D12/D3D12RenderContext.cpp
beldenfox: See my extended comments below. This is only necessary if we want to allow D3D clients to resize the back buffers to be smaller than the size of the HWND their attached to.

CMakeLists.txt
```diff
@@ -44,6 +44,17 @@ endmacro()
 macro(ADD_EXAMPLE_PROJECT TEST_NAME TEST_FILES LIB_FILES)
     if(APPLE)
         add_executable(${TEST_NAME} MACOSX_BUNDLE ${TEST_FILES})
+        set_target_properties(${TEST_NAME} PROPERTIES
```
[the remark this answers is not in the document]
beldenfox: High Resolution on a non-Retina Mac will be ignored. Well, there is no such thing as a non-Retina Mac. Any Mac can be attached to a high-resolution monitor which could invoke Retina mode. But, no, High Resolution is on by default in all Xcode-created projects and is benign when the screen is not Retina.

examples/Cpp/ExampleBase/ExampleBase.h
```diff
@@ -170,7 +170,7 @@ class ExampleBase
 
     ExampleBase(
         const std::wstring&     title,
-        const LLGL::Extent2D&   resolution  = { 800, 600 },
+        const LLGL::Extent2D&   windowSize  = { 800, 600 },
```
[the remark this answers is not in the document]
beldenfox: On the Mac a window that's 800x600 will typically have a drawable resolution of 1600x1200. One can argue whether the term 'windowSize' is appropriate (in the context of LLGL this is really the content size) but it's certainly not the resolution.

sources/Renderer/Direct3D11/D3D11RenderContext.cpp
```diff
@@ -222,7 +226,7 @@ void D3D11RenderContext::ResizeBackBuffer(const VideoModeDescriptor& videoModeDe
     backBuffer_.dsv.Reset();
 
     /* Resize swap-chain buffers, let DXGI find out the client area, and preserve buffer count and format */
-    auto hr = swapChain_->ResizeBuffers(0, 0, 0, DXGI_FORMAT_UNKNOWN, 0);
+    auto hr = swapChain_->ResizeBuffers(0, videoModeDesc.resolution.width, videoModeDesc.resolution.height, DXGI_FORMAT_UNKNOWN, 0);
```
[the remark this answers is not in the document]
beldenfox: See my extended comments. This is only necessary if we want to allow D3D clients to resize the back buffers to be smaller than the size of the HWND their attached to.

sources/Platform/MacOS/MacOSWindow.mm
```diff
@@ -144,7 +144,7 @@ - (NSSize)windowWillResize:(NSWindow*)sender toSize:(NSSize)frameSize
 - (void)windowDidResize:(NSNotification*)notification
 {
     //TODO: callback (here PostResize) must currently not be called while the NSEvent polling has not finished!
-    #if 0
+    #if 1
```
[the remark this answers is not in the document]
beldenfox: Not briefly. See my diatribe.

beldenfox:
I should have made it clearer that this pull request was really an invitation for a design review on SetVideoMode. I set aside the fullscreen logic to focus on two specific parts of the design.

The first issue is the recursive nature of SetVideoMode. It is being used to both invoke changes to the surface AND to respond to the resize events that those changes generate. This has some odd side-effects. For example, every time the window is resized the code tries to re-center it (the path is WM_SIZE => PostResize => SetVideoMode => Window::AdjustForVideoMode => Window::SetDesc with the ‘centered’ flag sent.) This only works on Windows because there’s a bug in Win32Window::SetDesc; it ignores the ‘centered’ flag unless the descriptor also changes the window’s size or position.

There are comments in Win32Window::SetDesc that suggest that SetDesc was generating WM_SIZE events that were leading back to SetDesc. This may have been aggravated by DirectX 12’s fullscreen implementation; entering fullscreen not only calls SetDesc it also leads to Direct3D’s SetFullscreenState call, both of which could generate more WM_SIZE events. At some point you tweaked the code to terminate the recursion by ensuring that SetDesc never sends out a PostResize message. I ran into that this weekend; I added some code that called SetDesc and the window resized without ever calling PostResize.

There are similar problems on the Mac. The code is trying to work around the re-centering problem by postponing the PostResize call until the user has stopped dragging. But when the user releases the mouse PostResize and SetVideoMode are finally called and try to re-center the window. Currently the Mac ignores the ‘centered’ flag and there’s a TODO comment about cleaning this up. In theory the Mac should also have recursion problems; a call to SetDesc will generate a resize event that will end up calling SetDesc again. I haven’t tested that but it’s in the design.

The cleanest way to fix all of this is to handle a resize event using code that updates the back buffers without touching the surface. As an experiment I pared down SetVideoMode so that’s all it does. That’s why I stubbed out fullscreen handling and Window::AdjustForVideoMode.

I also stubbed out fullscreen because there are even more recursion issues in the Direct3D backends. As of right now entering fullscreen using OnSetVideoMode in the Direct3D backends can generate resize events and those resize events are then handed back to OnSetVideoMode. I think that’s causing some issues but am still researching.

I will generate a pull request that tries to leave SetVideoMode alone as much as possible as a front-end call for going into and out of fullscreen mode. I will introduce a new back-end routine (e.g. RenderContext::ResizeBuffers or some such) just for handling WM_RESIZE events. I will try to keep the existing fullscreen logic but it’s complicated.

The second issue is support for arbitrarily sized back buffers. On the Mac I altered SetVideoMode to allow the back buffers to be sized independently of the surface size (which is necessary for Retina). Now the question is whether to carry the changes over to Direct3D as well. The use case on Windows is when an application wants to reduce the back buffer size because it’s running into fill rate issues in windowed mode on a high-resolution display (e.g. 4K). This would also be useful for apps that want to support borderless fullscreen windowed mode but run at a lower resolution than the display. In general this is a feature of DirectX with known use cases and I think LLGL should find a way to support it.

sources/Renderer/Direct3D11/D3D11RenderContext.cpp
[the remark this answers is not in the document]
LukasBanana: Are you saying this is a stability fix and/or performance improvement? It's more of a gut feeling that tells me it's safer to let DXGI figure out how large the back buffer must be. In this case, I'd prefer to have the D3D changes in a separate PR.

examples/Cpp/ExampleBase/ExampleBase.h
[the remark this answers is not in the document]
LukasBanana: I agree, if the resolution must be scaled the term "resolution" isn't perfect either. Can we move the discussion about the terminology to another PR? Maybe we can agree on `screenSize` or `contentSize`.

CMakeLists.txt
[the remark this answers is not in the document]
LukasBanana: I see. Then I wonder why CMake doesn't produce the corresponding property in the `.plist` file. But I'm fine with providing the proper file manually like this.

sources/Renderer/Direct3D12/D3D12RenderContext.cpp
LukasBanana: I'm thinking about unnecessary back buffer filtering. No scaling could potentially allow the simplest form of blitting the back buffer to the screen. With stretching, some form of sampling must be introduced which seems unnecessary for a back buffer that's supposed to have the exact same size as the client area.

sources/Renderer/Metal/MTRenderContext.mm
LukasBanana: Sounds good to me.

sources/Renderer/RenderContext.cpp
LukasBanana: The `Window` class is not necessary for an LLGL application as it merely adds a few more features like positioning and keyboard/mouse event handling so you can set up a prototype fairly easy. As soon as you want a more complex application you would ideally write your own implementation that derives from `Surface` as this class provides the minimum required interface for all backends. "View" is a reasonable term as well but more common in the macOS/iOS space. I think there are several parts in this PR we should split up for a separate discussion and approach.

sources/Renderer/RenderContext.cpp
LukasBanana: You're right, I see more and more games making use of a borderless window, too. If we can get this feature to work in LLGL, we can think about replacing the display configuration. That being said, I believe there are several philosophical design choices that are handled very differently between Mac and Windows. Most Windows games still provide the "true fullscreen" mode in which the user's display config is temporarily changed and I don't see the necessity (right now) to drop it immediately in LLGL. As for the Mac backend, this function was supposed to only move the App into fullscreen mode, without changing the display resolution or any other configuration, but iirc. it's not implemented yet.

Long story short, I'd prefer to handle the behavior of `SetDisplayMode` platform dependent. Whether it should be renamed then is another discussion.

LukasBanana:
Thanks for your detailed description, it makes a lot of things more clearer to me now. It sounds like you already did quite some investigation on that feature, so I'm happy to leave some of the new design choices up to you. You are right, the recursion in `SetVideoMode` is a weak design and it probably makes sense to split that into two functions.

sources/Renderer/RenderContext.cpp
beldenfox: I believe this will actively interfere with the way the Mac enters and exists borderless fullscreen mode. I think the facility should be present but called on by the platform code instead of the core code.

sources/Renderer/RenderContext.cpp
beldenfox: I believe both borderless fullscreen window and exclusive fullscreen are valid and LLGL should support both. Apple and Microsoft have written entire articles about how they've removed most of the bottlenecks related to borderless fullscreen mode. Still, I don't think LLGL can get rid of exclusive fullscreen so I'm thinking about how to structure the API to support both.

As for D3D, neither of the backends seem to be manipulating the display resolution using the Direct3D interfaces (e.g. ResizeTarget). Is that intentional? Bear in mind that I may be misunderstanding the Microsoft documentation.

sources/Renderer/OpenGL/Platform/MacOS/MacOSGLContext.mm
beldenfox: I'm having difficulty finding documentation on this function. The only write up is at the following URL and it basically boils down to "call these routines to change the back buffer sizes for OpenGL". https://developer.apple.com/library/archive/documentation/GraphicsImaging/Conceptual/OpenGL-MacProgGuide/opengl_contexts/opengl_contexts.html#//apple_ref/doc/uid/TP40001987-CH216-SW28

sources/Renderer/Direct3D11/D3D11RenderContext.cpp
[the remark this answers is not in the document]
beldenfox: It's a feature enhancement. If we want a Windows client to be able to downsize the the back buffers for performance reasons than we have to pass in explicit sizes to ResizeBuffers and enable stretch scaling mode so the smaller back buffer can stretch across the entire surface. Does this incur a performance cost if it's not necessary? I don't know and Microsoft isn't really saying.

sources/Renderer/Direct3D12/D3D12RenderContext.cpp
beldenfox: Again, this allows a client to use a back buffer that's smaller than the HWND but still fills the HWND. The Microsoft documentation doesn't make clear whether turning this on imposes a performance penalty when the HWND and back buffer are the same size and there's no need to stretch. Stretch seems to be the default on D3D11 so I'm guessing it does not.

LukasBanana:
My apologies that I never took the time to do the design review you asked for, but I don't actively work on this project anymore. If there is no update from your side, I'm going to close this PR.

Issue #82, opened 2023-04-07
Support MSYS2 MINGW64 on Windows


iahung:
On MSYS2, there is no Direct3D 11 or Direct3D 12, but there is Vulkan. OpenGL and Vulkan renders will available.

LukasBanana:
Is this a question? I am not familiar with MSYS2, but I'm open for suggestions how to extend `CMakeLists.txt` script to support it.

PR #83, opened 2023-04-19
Build on MSYS2 MINGW64

iahung:
https://github.com/LukasBanana/LLGL/issues/82

### DIFF
--- a/BuildMsys2.sh
+++ b/BuildMsys2.sh
@@ -1,0 +1,39 @@
+#!/bin/sh
+
+SOURCE_DIR=$PWD
+OUTPUT_DIR="build_msys2"
+
+# Ensure we are inside the repository folder
+if [ ! -f "CMakeLists.txt" ]; then
+    echo "error: file not found: CMakeLists.txt"
+    exit 1
+fi
+
+# Make output build folder
+if [ "$#" -eq 1 ]; then
+    OUTPUT_DIR=$1
+else
+    if [ ! "$#" -eq 0 ]; then
+        echo "error: too many arguemnts"
+        echo "usage: BuildMsys2.sh [OUTPUT_DIR]"
+        exit 1
+    fi
+fi
+
+if [ ! -d "$OUTPUT_DIR" ]; then
+    mkdir "$OUTPUT_DIR"
+fi
+
+cd "$OUTPUT_DIR"
+
+# Checkout external depenencies
+GAUSSIAN_LIB_DIR="GaussianLib/include"
+
+if [ ! -d "$GAUSSIAN_LIB_DIR" ]; then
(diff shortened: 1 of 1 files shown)


BuildMsys2.sh
```diff
@@ -0,0 +1,39 @@
...
+    exit 1
+fi
+
+# Make output build folder
+if [ "$#" -eq 1 ]; then
+    OUTPUT_DIR=$1
+else
+    if [ ! "$#" -eq 0 ]; then
+        echo "error: too many arguemnts"
+        echo "usage: BuildLinux.sh [OUTPUT_DIR]"
```
LukasBanana, line 18: `BuildLinux.sh` should probably be renamed to `BuildMsys2.sh` here.

BuildMsys2.sh
```diff
@@ -0,0 +1,39 @@
...
+
+# Checkout external depenencies
+GAUSSIAN_LIB_DIR="GaussianLib/include"
+
+if [ ! -d "$GAUSSIAN_LIB_DIR" ]; then
+    git clone https://github.com/LukasBanana/GaussianLib.git
+fi
+
+# Build into output directory
+cmake -DLLGL_BUILD_RENDERER_OPENGL=ON -DLLGL_BUILD_RENDERER_VULKAN=ON -DLLGL_BUILD_RENDERER_DIRECT3D11=OFF -DLLGL_BUILD_RENDERER_DIRECT3D12=OFF -DLLGL_BUILD_EXAMPLES=ON -DGaussLib_INCLUDE_DIR:STRING="$OUTPUT_DIR/$GAUSSIAN_LIB_DIR" -DCMAKE_BUILD_TYPE=Release -S "$SOURCE_DIR"
```
LukasBanana, line 37: Does MSYS2 come with the necessary Vulkan SDK out of the box since `LLGL_BUILD_RENDERER_VULKAN` is enabled here?

BuildMsys2.sh
iahung: Thank you.

BuildMsys2.sh
LukasBanana: So I presume the user would see an error by default and has to resolve it in order to continue with the build process. I disabled Vulkan in the `BuildLinux.sh` script for the same reason. So I'm wondering if we should only use the GL backend by default for MSYS2 as well, to simplify that build process. If someone is interested in Vulkan, they only have to enable it in the script manually, but it doesn't interrupt other's workflow.

Could you please also squash your commits? The rest looks good to me.

iahung: I did everything as you said.

LukasBanana:
Looks good. Thanks for your contribution.

PR #90, opened 2023-08-15
Include cstdint to fix build on some platforms

bsaleil:
Add cstdint include to fix the build on some platforms. I wasn't able to build LLGL without that include on my Linux machine.

### DIFF
--- a/sources/Renderer/Null/RenderState/NullFence.h
+++ b/sources/Renderer/Null/RenderState/NullFence.h
@@ -12,7 +12,7 @@
 #include <LLGL/Fence.h>
 #include <string>
 #include <atomic>
-
+#include <cstdint>
 
 namespace LLGL
 {


LukasBanana:
Thanks for the fix. I am curious why this wasn't picked up by the CIS build for Ubuntu. What is your build configuration (GCC or Clang)? Perhaps a different compiler toolchain?

bsaleil:
Yeah that's weird. I had the same issue with both gcc 13 and clang 15. Maybe the CIS is using and older version of gcc that was using `<cstdint>` more widely in its headers ? (https://gcc.gnu.org/gcc-13/porting_to.html)

LukasBanana:
FWIW: The [CIS build](https://github.com/LukasBanana/LLGL/actions/workflows/ci_linux.yml) is using GNU 11.4.0.

PR #94, opened 2023-08-27
Add support for Shader Model 6.0 and above (DXC) to D3D12Shader

LukasBanana:
I opened this PR for discussion and extending it to make this feature optional, i.e. the Windows SDK should not have to be installed. I also opened this PR into the new `staged` branch for integrating larger PRs.


sources/Renderer/DXCommon/DXCore.cpp
```diff
@@ -15,6 +15,7 @@
 #include <stdexcept>
 #include <algorithm>
 #include <d3dcompiler.h>
+#include <dxcapi.h>
```
LukasBanana, line 18: I presume this header is included in the Windows SDK, so let me add a new CMake option to exclude it from builds that don't have that SDK installed. Same goes for the code that uses it. Eventually, I will probably move it out into a separate file to have build specific code nicely separated instead of having several `#ifdef` preprocessor blocks.

sources/Renderer/Direct3D12/D3D12RenderSystem.cpp
```diff
@@ -527,6 +533,20 @@ void D3D12RenderSystem::CreateDevice()
     }
 }
 
+void D3D12RenderSystem::CreateDxcInterface()
+{
+    dxcModule_ = LoadLibrary(L"dxcompiler.dll");
```
LukasBanana, line 538: This one should use the `TEXT()` macro instead of explicit wide character string literals in case someone disables the `UNICODE` macro on Windows.

sources/Renderer/Direct3D12/Shader/D3D12Shader.cpp
```diff
@@ -232,6 +235,30 @@ void D3D12Shader::BuildStreamOutput(UINT numVertexAttribs, const VertexAttribute
...
+bool IsProfileDxcAppropriate(const char* target)
+{
+    size_t szLen = StrLength(target);
+
+    // LLGL allows for a blank string to be sent into the target of a ShaderDescriptor, but
+    // FXC nor DXC supports this behavior, so it doesn't matter if we send to FXC or DXC.
+    if (szLen == 0)
+        return false;
+
+    for_range(i, szLen)
```
LukasBanana, line 247: `StrLength` relies on the string being NUL-terminated, just like `strlen`, so this loop could combine the search for the end and the `_` character. Perhaps something like this including the explanation of what we are parsing:
```cpp
// Search for first occurrence of '_' to parse the FXC/DXC SM pattern '[lib|vs|ps|...]_D_D', e.g. 'vs_6_0'
for (const char* c = target; c != '\0'; ++c)
{
    if (*c == '_')
    {
        // Safe to use next character in NUL-terminated string, since it's either our major version number or '\0'
        char majorShaderVersion = c[1];
        if (majorShaderVersion >= '6')
            return true;
    }
}
```

sources/Renderer/DXCommon/DXCore.cpp
gdianaty: It is included with the Windows Kits, like `d3dcompiler.h`, and it should exist alongside any copy of the d3d12 headers.

sources/Renderer/Direct3D12/D3D12RenderSystem.cpp
gdianaty: Gotchya! I was wondering about that - I'll tweak that here in a moment.

sources/Renderer/DXCommon/DXCore.cpp
LukasBanana: I believe `d3dcompiler.h` and `d3d12.h` are included with Visual Studio, but maybe that changed with newer Visual Studio editions; I am still working with VS 2015.

gdianaty: Given the relative newness of SM6 and the DXC, it sounds like you're right - it's changed as of a recent version of the Windows SDK. Are you unable to build, or is the dependency missing?

As a note, the header for the DXC is available at the repository for it ([here](https://github.com/microsoft/DirectXShaderCompiler)) and should work unconditional of SDK version or compiler, if included as a submodule.

LukasBanana: I see the problem now, my CMake version does not select the latest Windows SDK I have installed on my machine. I can work that around by manually selecting it in VS for now.

LukasBanana:
I got a bit caught up in the different branches but I committed your changes with c5755c3.

gdianaty:
Excellent! Very good to see.

PR #95, opened 2023-08-28
Move D3D builtin shaders from being DLL resources to being bytecode objects.

gdianaty:
# The Problem
As of now, when LLGL is built as static (`LLGL_BUILD_STATIC_LIB`) it is unable to perform any mip computations. This is because LLGL's mip generation functionality uses a compute shader to achieve mip generation.

This is because the Windows Resource Compiler is not honored by the MSVC Librarian and is therefore lost at compile-time. Static versions of the library don't have access to the shader bytecode.

# The (Proposed) Solution
Move shader bytecode to a header (`/Fh`). This way it's included in static versions of the library. With this edit made, all examples run as expected and mip generation occurs as if the library was included as a dynamic library.


LukasBanana (commented):
Looks good to me overall. I thought I fixed that by using `GetModuleHandle(nullptr)` when locating a resource from a static build:
https://github.com/LukasBanana/LLGL/blob/48da88a5a441574dfc928fc513b36cdab2d8e6a5/sources/Renderer/DXCommon/DXCore.cpp#L52
But I guess I didn't test that properly.
Let me add some suggestions for a few modifications.

sources/Renderer/Direct3D11/Shader/Builtin/Compile.bat
```diff
@@ -1,10 +1,10 @@
 REM Compiles all shaders with fxc
 REM Use VisualStudio command line tools to run this batch script
 
-fxc /D TEXTURE_DIM=1 /T cs_5_0 /E CopyTextureFromBuffer /Fo CopyTextureFromBufferCS.Dim1D.cso CopyTextureFromBuffer.hlsl
-fxc /D TEXTURE_DIM=2 /T cs_5_0 /E CopyTextureFromBuffer /Fo CopyTextureFromBufferCS.Dim2D.cso CopyTextureFromBuffer.hlsl
-fxc /D TEXTURE_DIM=3 /T cs_5_0 /E CopyTextureFromBuffer /Fo CopyTextureFromBufferCS.Dim3D.cso CopyTextureFromBuffer.hlsl
+fxc /D TEXTURE_DIM=1 /T cs_5_0 /E CopyTextureFromBuffer /Fh CopyTextureFromBufferCS.Dim1D.inc /Vn CopyTextureFromBufferCS_Dim1D CopyTextureFromBuffer.hlsl
```
LukasBanana, line 4: At first I thought we could use LLGL's `ReadFileAsHexStringDecl` script to generate the same format for the header files as for the Metal backend, but I guess using FXC's only capability to generate a C header is just fine and convenient.

I would like to have the same naming convention here for the extension, though. So changing this from `.inc` to `.cso.inl` would make it clear it's an inline-only header (with no header guard) for a CSO (Compute Shader Object).

Speaking of naming convention: FXC actually generates the default name in the same fashion as I have my global variables/constants with the `g_` prefix, but since FXC only uses the entry point name, we do in fact have to override the output name. Just adding a `g_` prefix to the output name (for `/Vn`) would follow the same pattern.

Could you please also add `/Qstrip_reflect` to the FXC command? I didn't realize how much bigger those resource files become but the text format made it more obvious.

LukasBanana (changes requested):

sources/Renderer/Direct3D11/Shader/Builtin/Compile.bat
gdianaty: Whammy! Got those suggestions implemented real fast.

LukasBanana:
Thanks for fixing this 👍

PR #98, opened 2023-10-01
Add field for source name functionality in HLSL to LLGL::ShaderDescriptor and fix D3D include pathing.

gdianaty:
# The Problem
When LLGL's built-in D3D shader compilers are running and an #include directive is used, that directive is usually unable to be completed due to the shader being transformed in such a way that makes the D3D compiler unable to extrapolate include paths.

# The (Proposed) Solution
Allow LLGL::ShaderDescriptor to hold a field for source names and pass that down to the underlying D3D compiler implementation as-needed.


sources/Renderer/Direct3D11/Shader/D3D11Shader.cpp
```diff
@@ -156,17 +156,22 @@ bool D3D11Shader::CompileSource(ID3D11Device* device, const ShaderDescriptor& sh
...
     SIZE_T      sourceLength    = 0;
+    const char* sourceName      = nullptr;
 
     if (shaderDesc.sourceType == ShaderSourceType::CodeFile)
     {
         fileContent     = ReadFileString(shaderDesc.source);
         sourceCode      = fileContent.c_str();
         sourceLength    = fileContent.size();
+
+        if (shaderDesc.source != nullptr)
```
LukasBanana, line 167: I think it would be good to always prefer `shaderDesc.name` if it's specified and only use `shaderDesc.source` as a fallback case. The rest looks good to me.

PR #104, opened 2024-02-24
Fix incorrect sampler settings used in ResourceBinding Example.cpp

duckdoom4:


LukasBanana:
That's indeed the wrong sampler setting but neither of those samplers are used in the example. They can both be removed since the PSO layout uses static samplers. There are enough other examples to show how to create dynamic sampler states.

PR #111, opened 2024-03-29
[Draft] [D3D] Support variable refresh rate displays (vsync off) in borderless windowed

duckdoom4:
I made an attempt to improve support for vsync off in borderless windowed mode. (See https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/variable-refresh-rate-displays)

The [documentation recommends](https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/for-best-performance--use-dxgi-flip-model) we no longer use `DXGI_SWAP_EFFECT_DISCARD` and switch to `DXGI_SWAP_EFFECT_FLIP_*` instead. 

Still TODO:
- [ ] MSAA is not supported in this model. The dx11 implementation will need to be updated to support this. There is a detailed explaination on how to resolve that [here](https://learn.microsoft.com/en-us/windows/win32/api/dxgi/ne-dxgi-dxgi_swap_effect#remarks). (Would love your help with this one, I assume we can use an almost exact copy of the dx12 implementation?)
- [x] Disable tearing if fullscreen exclusive mode is active
- [x] Verify the code will work under different configurations
- [ ] Update the dx11 code so it matches the dx12 version? (Aka. use device to create swap chain, and move tearing support check to device as well)

Let me know what you think, would love your feedback :)

### DIFF
--- a/sources/Renderer/Direct3D11/D3D11RenderSystem.cpp
+++ b/sources/Renderer/Direct3D11/D3D11RenderSystem.cpp
@@ -551,8 +551,25 @@ void D3D11RenderSystem::ClearStateForAllContexts()
 void D3D11RenderSystem::CreateFactory()
 {
     /* Create DXGI factory */
-    HRESULT hr = CreateDXGIFactory(IID_PPV_ARGS(factory_.ReleaseAndGetAddressOf()));
+#if LLGL_D3D11_ENABLE_FEATURELEVEL >= 2
+
+    HRESULT hr = CreateDXGIFactory2(0, IID_PPV_ARGS(&factory2_));
+    DXThrowIfCreateFailed(hr, "IDXGIFactory2");
+    factory2_.As(&factory_);
+    factory2_.As(&factory1_);
+
+#elif LLGL_D3D11_ENABLE_FEATURELEVEL >= 1
+
+    HRESULT hr = CreateDXGIFactory1(IID_PPV_ARGS(&factory1_));
+    DXThrowIfCreateFailed(hr, "IDXGIFactory1");
+    factory1_.As(&factory_);
+
+#else
+
+    HRESULT hr = CreateDXGIFactory(IID_PPV_ARGS(&factory_));
     DXThrowIfCreateFailed(hr, "IDXGIFactory");
+
+#endif
 }
 
 void D3D11RenderSystem::QueryVideoAdapters(long flags, ComPtr<IDXGIAdapter>& outPreferredAdatper)
--- a/sources/Renderer/Direct3D11/D3D11RenderSystem.cpp
+++ b/sources/Renderer/Direct3D11/D3D11RenderSystem.cpp
@@ -551,25 +551,29 @@ void D3D11RenderSystem::ClearStateForAllContexts()
 void D3D11RenderSystem::CreateFactory()
 {
     /* Create DXGI factory */
+    HRESULT hr = S_OK;
+
 #if LLGL_D3D11_ENABLE_FEATURELEVEL >= 2
-
-    HRESULT hr = CreateDXGIFactory2(0, IID_PPV_ARGS(&factory2_));
-    DXThrowIfCreateFailed(hr, "IDXGIFactory2");
-    factory2_.As(&factory_);
-    factory2_.As(&factory1_);
-
-#elif LLGL_D3D11_ENABLE_FEATURELEVEL >= 1
-
-    HRESULT hr = CreateDXGIFactory1(IID_PPV_ARGS(&factory1_));
-    DXThrowIfCreateFailed(hr, "IDXGIFactory1");
-    factory1_.As(&factory_);
-
-#else
-
-    HRESULT hr = CreateDXGIFactory(IID_PPV_ARGS(&factory_));
-    DXThrowIfCreateFailed(hr, "IDXGIFactory");
-
+    hr = CreateDXGIFactory2(0, IID_PPV_ARGS(&factory2_));
+    if (SUCCEEDED(hr))
+    {
+        factory2_.As(&factory_);
+        factory2_.As(&factory1_);
+        return;
+    }
 #endif
+
+#if LLGL_D3D11_ENABLE_FEATURELEVEL >= 1
+    hr = CreateDXGIFactory1(IID_PPV_ARGS(&factory1_));
+    if (SUCCEEDED(hr))
+    {
+        factory1_.As(&factory_);
+        return;
+    }
+#endif
+
+    hr = CreateDXGIFactory(IID_PPV_ARGS(&factory_));
+    DXThrowIfCreateFailed(hr, "IDXGIFactory");
 }
 
 void D3D11RenderSystem::QueryVideoAdapters(long flags, ComPtr<IDXGIAdapter>& outPreferredAdatper)
@@ -688,9 +692,10 @@ void D3D11RenderSystem::QueryDXDeviceVersion()
 {
     LLGL_ASSERT_PTR(device_);
 
+    HRESULT hr = S_OK;
     /* Try to get an extended D3D11 device */
     #if LLGL_D3D11_ENABLE_FEATURELEVEL >= 3
-    HRESULT hr = device_->QueryInterface(IID_PPV_ARGS(&device3_));
+    hr = device_->QueryInterface(IID_PPV_ARGS(&device3_));
     if (FAILED(hr))
     #endif
     {
--- a/sources/Renderer/Direct3D11/D3D11SwapChain.cpp
+++ b/sources/Renderer/Direct3D11/D3D11SwapChain.cpp
@@ -10,6 +10,7 @@
 #include "D3D11ObjectUtils.h"
 #include "../DXCommon/DXTypes.h"
 #include "../../Core/CoreUtils.h"
+#include "../../Core/CompilerExtensions.h"
 #include <LLGL/Platform/NativeHandle.h>
 #include <LLGL/Log.h>
 
@@ -268,17 +269,17 @@ void D3D11SwapChain::CreateSwapChain(IDXGIFactory* factory, const Extent2D& reso
 {
     /* Pick and store color format */
     colorFormat_ = DXGI_FORMAT_R8G8B8A8_UNORM;//DXGI_FORMAT_B8G8R8A8_UNORM
-    /* Clamp buffer count between 1 and max buffers */
-    swapBuffers = std::max(1u, std::min<std::uint32_t>(swapBuffers, DXGI_MAX_SWAP_CHAIN_BUFFERS));
-
-    /* Find suitable multi-samples for color format */
-    swapChainSampleDesc_ = D3D11RenderSystem::FindSuitableSampleDesc(device_.Get(), colorFormat_, 1); // TODO: resolve multi-sampling
-
+    
     /* Create swap chain for window handle */
     NativeHandle wndHandle = {};
     GetSurface().GetNativeHandle(&wndHandle, sizeof(wndHandle));
 
 #if LLGL_D3D11_ENABLE_FEATURELEVEL >= 3
+    /* Clamp buffer count between 2 and max buffers */
+    swapBuffers = std::max(2u, std::min<std::uint32_t>(swapBuffers, DXGI_MAX_SWAP_CHAIN_BUFFERS));
+
+    /* Find suitable multi-samples for color format */
+    swapChainSampleDesc_ = D3D11RenderSystem::FindSuitableSampleDesc(device_.Get(), colorFormat_, 1); // TODO: resolve multi-sampling
 
     DXGI_SWAP_CHAIN_DESC1 swapChainDesc = {};
     {
@@ -288,7 +289,7 @@ void D3D11SwapChain::CreateSwapChain(IDXGIFactory* factory, const Extent2D& reso
             swapChainDesc.SampleDesc    = swapChainSampleDesc_;
             swapChainDesc.BufferUsage   = DXGI_USAGE_RENDER_TARGET_OUTPUT;
             swapChainDesc.BufferCount   = swapBuffers;
-            swapChainDesc.SwapEffect    = DXGI_SWAP_EFFECT_FLIP_DISCARD; // TODO: requires BufferCount >= 2 && SampleDesc.Count == 1
+            swapChainDesc.SwapEffect    = DXGI_SWAP_EFFECT_FLIP_DISCARD; // FLIP effect requires BufferCount >= 2 && SampleDesc.Count == 1
             swapChainDesc.Flags         = (tearingSupported_ ? DXGI_SWAP_CHAIN_FLAG_ALLOW_TEARING : 0u);
     }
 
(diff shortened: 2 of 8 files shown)


LukasBanana (changes requested):
I'm happy to see this project getting modernized with features like variable refresh rate.

With respect to C++17/C++20 features: I want to keep compatibility strictly with C++11. My rationale behind this is to stay compatible with as many platforms and legacy systems as possible. Supporting legacy systems is basically only intended for the GL backend, but I like to have a cohesive code base that uses the same language standard throughout. Having said that, I don't mind adding some of those features like `[[maybe_unused]]` to the [CompilerExtensions.h](https://github.com/LukasBanana/LLGL/blob/master/sources/Core/CompilerExtensions.h) file like so:
```cpp
#if __cplusplus >= 202002L // C++20
#   define LLGL_MAYBE_UNUSED [[maybe_unused]]
#else
#   define LLGL_MAYBE_UNUSED
#endif
```
With the macro that's already defined there I made an exception with my naming guideline so you could write `if unlikely(X) {}`.

sources/Renderer/Direct3D12/D3D12SwapChain.cpp
```diff
@@ -103,7 +104,10 @@ void D3D12SwapChain::SetDebugName(const char* name)
 void D3D12SwapChain::Present()
 {
     /* Present swap-chain with vsync interval */
-    HRESULT hr = swapChainDXGI_->Present(syncInterval_, 0);
+    bool tearingEnabled = tearingSupported_ && syncInterval_ == 0;
```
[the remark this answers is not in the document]
duckdoom4: Ah, i thought dx12 didn't support the same full screen exclusive mode as dx11. But if you did implement that, then yes, it would need the same check

sources/Renderer/Direct3D11/D3D11SwapChain.cpp
```diff
@@ -71,7 +75,9 @@ void D3D11SwapChain::SetDebugName(const char* name)
 
 void D3D11SwapChain::Present()
 {
-    swapChain_->Present(swapChainInterval_, 0);
+    bool tearingEnabled = tearingSupported_ && swapChainInterval_ == 0; // TODO: disable if fullscreen exclusive mode is enabled
```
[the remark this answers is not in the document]
duckdoom4: Yeah, I think so. The documentation didn't mention a way to do this so I'd need to do some googling to see if there is some (easy) way to check

sources/Renderer/Direct3D11/D3D11SwapChain.cpp
```diff
@@ -260,19 +266,45 @@ static UINT GetPrimaryDisplayRefreshRate()
...
-    const DXGI_RATIONAL refreshRate{ GetPrimaryDisplayRefreshRate(), 1 };
-
     /* Pick and store color format */
     colorFormat_ = DXGI_FORMAT_R8G8B8A8_UNORM;//DXGI_FORMAT_B8G8R8A8_UNORM
+    /* Clamp buffer count between 1 and max buffers */
+    swapBuffers = std::max(1u, std::min<std::uint32_t>(swapBuffers, DXGI_MAX_SWAP_CHAIN_BUFFERS));
 
     /* Find suitable multi-samples for color format */
-    swapChainSampleDesc_ = D3D11RenderSystem::FindSuitableSampleDesc(device_.Get(), colorFormat_, samples);
+    swapChainSampleDesc_ = D3D11RenderSystem::FindSuitableSampleDesc(device_.Get(), colorFormat_, 1); // TODO: resolve multi-sampling
```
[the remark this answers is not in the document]
duckdoom4: Sounds good. I think I'll have some time on Monday so I can finish it then

sources/Renderer/Direct3D11/D3D11SwapChain.cpp
```diff
@@ -381,6 +418,23 @@ void D3D11SwapChain::RestoreDebugNames(const std::string (&debugNames)[4])
     }
 }
 
+void D3D11SwapChain::CheckTearingSupport([[maybe_unused]] IDXGIFactory* factory)
```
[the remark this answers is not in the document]
duckdoom4: Ah right, I forgot to put it back in c++11 mode after I made the MR to fix the c++20 issues I had in my own project. I'll put it back so I don't forget in the future

sources/Renderer/Direct3D11/D3D11SwapChain.cpp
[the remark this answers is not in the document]
duckdoom4: I checked out this example they have on github, which also contains the tearing support. https://github.com/microsoft/DirectX-Graphics-Samples/blob/master/Samples/Desktop/D3D12Fullscreen/src/D3D12Fullscreen.cpp

It seems all they do there is test `GetFullScreenState` in the swapchain resize function and store that in the `m_windowedMode` bool. I can do the same here. 

They also have a separate handler for alt+enter replacing it with spacebar, but if you look at the window message that handles the space press the comment suggests that alt+enter does the same thing. So I guess that's not a requirement. (Also didn't see any mention of this in any other documentation so far)

LukasBanana:
Changes look good at first glance. I'll take a closer look and test this on Thursday.

sources/Renderer/Direct3D12/D3D12SwapChain.cpp
```diff
@@ -441,13 +448,17 @@ HRESULT D3D12SwapChain::CreateResolutionDependentResources(const Extent2D& resol
...
             swapChainDesc.AlphaMode             = DXGI_ALPHA_MODE_IGNORE;
-            swapChainDesc.Flags                 = 0;
+            swapChainDesc.Flags                 = (tearingSupported_ ? DXGI_SWAP_CHAIN_FLAG_ALLOW_TEARING : 0u);
         }
         auto swapChain = renderSystem_.CreateDXSwapChain(swapChainDesc, wndHandle.window);
 
         swapChain.As(&swapChainDXGI_);
     }
 
+    BOOL fullscreenState;
```
LukasBanana, line 458: Unless it's a larger buffer that is intended to be uninitialized at allocation time (like `LLGL::DynamicByteArray{ N, LLGL::UninitializeTag{} }` for instance), I'd prefer to have all variables default initialized. Skipping the `= FALSE` here is really just a micro optimization.

sources/Renderer/Direct3D12/D3D12RenderSystem.cpp
```diff
@@ -19,19 +19,19 @@
...
 
 #include "Texture/D3D12MipGenerator.h"
 
-#include "RenderState/D3D12GraphicsPSO.h"
 #include "RenderState/D3D12ComputePSO.h"
+#include "RenderState/D3D12GraphicsPSO.h"
 
 #include <LLGL/Backend/Direct3D12/NativeHandle.h>
 
+#include <dxgi1_5.h>
```
LukasBanana, line 34: What minimum Windows SDK version requires DXGI 1.5? So far DXGI 1.4 was the highest used DXGI version in the D3D12 backend, which I set the WinSDK version to 10.0.10240.0 in the [docs page](https://github.com/LukasBanana/LLGL/tree/master/docu#windows). Would this require to lift the SDK version or can we wrap this into an `#if LLGL_D3D12_ENABLE_FEATURELEVEL` block like with the D3D11 backend? Having that said, I think I'll have to split this macro up anyway, since feature level, D3D version, DXGI version, and shader-model are all separate concepts that are not tied together 1:1, but right now all of these things are wrapped under the same CMake option and macro.

sources/Renderer/Direct3D12/D3D12RenderSystem.cpp
```diff
@@ -19,19 +19,19 @@
...
 
 #include "Buffer/D3D12Buffer.h"
 #include "Buffer/D3D12BufferArray.h"
 #include "Buffer/D3D12BufferConstantsPool.h"
 
 #include "Texture/D3D12MipGenerator.h"
 
-#include "RenderState/D3D12GraphicsPSO.h"
 #include "RenderState/D3D12ComputePSO.h"
+#include "RenderState/D3D12GraphicsPSO.h"
```
LukasBanana, line 30: I presume this change isn't really necessary.

sources/Renderer/Direct3D11/D3D11SwapChain.cpp
```diff
@@ -260,18 +267,49 @@ static UINT GetPrimaryDisplayRefreshRate()
...
+            swapChainDesc.Format        = colorFormat_;
+            swapChainDesc.SampleDesc    = swapChainSampleDesc_;
+            swapChainDesc.BufferUsage   = DXGI_USAGE_RENDER_TARGET_OUTPUT;
+            swapChainDesc.BufferCount   = swapBuffers;
+            swapChainDesc.SwapEffect    = DXGI_SWAP_EFFECT_FLIP_DISCARD; // FLIP effect requires BufferCount >= 2 && SampleDesc.Count == 1
+            swapChainDesc.Flags         = (tearingSupported_ ? DXGI_SWAP_CHAIN_FLAG_ALLOW_TEARING : 0u);
+    }
+
+    ComPtr<IDXGIFactory2> factory2;
+    HRESULT hr = factory->QueryInterface(IID_PPV_ARGS(&factory2));
```
LukasBanana, line 297: Just for the factory query, I think it makes sense to move this upward within this `#if` block and if it fails, fallback to the alternative swap-chain creation, just like you did in `D3D11SwapChain::CheckTearingSupport`. If `CreateSwapChainForHwnd` fails with a valid factory, we should indeed crash with a callstack.

sources/Renderer/Direct3D11/D3D11SwapChain.cpp
```diff
@@ -260,18 +267,49 @@ static UINT GetPrimaryDisplayRefreshRate()
...
+    /* Create swap chain for window handle */
+    NativeHandle wndHandle = {};
+    GetSurface().GetNativeHandle(&wndHandle, sizeof(wndHandle));
+
+#if LLGL_D3D11_ENABLE_FEATURELEVEL >= 3
+    /* Clamp buffer count between 2 and max buffers */
+    swapBuffers = std::max(2u, std::min<std::uint32_t>(swapBuffers, DXGI_MAX_SWAP_CHAIN_BUFFERS));
+
+    /* Find suitable multi-samples for color format */
+    swapChainSampleDesc_ = D3D11RenderSystem::FindSuitableSampleDesc(device_.Get(), colorFormat_, 1); // TODO: resolve multi-sampling
```
LukasBanana, line 282: I can take care of adopting the same back-buffer creation as in D3D12.

sources/Renderer/Direct3D11/D3D11RenderSystem.cpp
```diff
@@ -551,8 +551,25 @@ void D3D11RenderSystem::ClearStateForAllContexts()
 void D3D11RenderSystem::CreateFactory()
 {
     /* Create DXGI factory */
-    HRESULT hr = CreateDXGIFactory(IID_PPV_ARGS(factory_.ReleaseAndGetAddressOf()));
+#if LLGL_D3D11_ENABLE_FEATURELEVEL >= 2
+
+    HRESULT hr = CreateDXGIFactory2(0, IID_PPV_ARGS(&factory2_));
```
LukasBanana, line 556: These should also all fallback to the next lower factory version if they fail. `LLGL_D3D11_ENABLE_FEATURELEVEL` is only a guard to let LLGL compile on systems with older Windows SDKs. It shouldn't require that the runtime version supports anything higher than Direct3D 11.0.

LukasBanana (changes requested):
I left a few comments. With the SDK requirement for DXGI 1.5, I'm okay with leaving that as-is since the solution with the `LLGL_D3D12_ENABLE_FEATURELEVEL` macro is not optimal right now anyway. If you can adjust the other points like making sure we can fall back to lower versions of the factory if the runtime doesn't support it (regardless of what's enabled via the feature level macros) I'll merge this into the master branch and adjust the D3D11 backend to support multi-sampled backbuffers.

sources/Renderer/Direct3D11/D3D11SwapChain.cpp
duckdoom4: Makes sense, will do that

sources/Renderer/Direct3D12/D3D12RenderSystem.cpp
duckdoom4: True, that was my auto formatter sorting the includes

sources/Renderer/Direct3D12/D3D12RenderSystem.cpp
duckdoom4: According to [this page](https://github.com/toivjon/dxgi-sandbox), it's Windows 10.

sources/Renderer/Direct3D12/D3D12SwapChain.cpp
duckdoom4: Ah yeah missed that, my bad

duckdoom4:
Had a bussy week, but got the comments addressed now.

Added the fallback to the next lower factory version on failure, and added the fallback to the next swap chain version on failure by adding a new function to make it more readable

sources/Renderer/Direct3D11/D3D11RenderSystem.cpp
```diff
@@ -671,9 +692,10 @@ void D3D11RenderSystem::QueryDXDeviceVersion()
 {
     LLGL_ASSERT_PTR(device_);
 
+    HRESULT hr;
```
duckdoom4, line 695: Moved this here, because line 681/703 would fail to compile if `LLGL_D3D11_ENABLE_FEATURELEVEL`  is set to `2`

sources/Renderer/Direct3D11/D3D11RenderSystem.cpp
```diff
@@ -671,9 +692,10 @@ void D3D11RenderSystem::QueryDXDeviceVersion()
 {
     LLGL_ASSERT_PTR(device_);
 
+    HRESULT hr = S_OK;
```
duckdoom4, line 695: Moved this here, because line 681/703 would fail to compile if LLGL_D3D11_ENABLE_FEATURELEVEL is set to 2

duckdoom4, line 695: Moved this here, because line 681/703 would fail to compile if LLGL_D3D11_ENABLE_FEATURELEVEL is set to 2

sources/Renderer/Direct3D11/D3D11SwapChain.cpp
```diff
@@ -286,9 +305,45 @@ void D3D11SwapChain::CreateSwapChain(IDXGIFactory* factory, const Extent2D& reso
         swapChainDesc.Windowed                  = TRUE;//(fullscreen ? FALSE : TRUE);
         swapChainDesc.SwapEffect                = DXGI_SWAP_EFFECT_DISCARD;
     }
-    HRESULT hr = factory->CreateSwapChain(device_.Get(), &swapChainDesc, swapChain_.ReleaseAndGetAddressOf());
+
+    hr = factory->CreateSwapChain(device_.Get(), &swapChainDesc, swapChain_.ReleaseAndGetAddressOf());
```
LukasBanana, line 309: This one fails with `'hr': undeclared identifier` if `LLGL_D3D11_ENABLE_FEATURELEVEL` is 2 or lower (see [AppVeyor CIS job](https://ci.appveyor.com/project/LukasBanana/llgl/builds/49605941/job/3undo0l8qxus4w7e#L257)). Can likely be fixed the same way you fixed it in _D3D11RenderSystem.cpp_.

LukasBanana (changes requested):
Just one compile error to fix and we're good to go. I'm also a bit slow on responding right now, because I'm on vacation and will be back on Thursday.

LukasBanana (approved):
Looks good. Thanks again for your contribution. I will add the MSAA support for D3D11 next weekend.

duckdoom4:
Yeah you're welcome. Happy to help. I'm actually looking at adding some other features as well so might make some more in the near future.

For example adding support for HDR framebuffer output for HDR displays.

Also I need bindless textures for a research project I'm working on. 
Does LLGL support that currently? I already read that dx11 does not support that,since it's a shader model 5.1 thing

LukasBanana:
Added support for multi-sampled back-buffers with the FLIP swap-effect with 5778b34.

PR #113, opened 2024-05-22
Fix crash in D3D12 when using LLGL::RenderingDebugger.

gdianaty:
Caught this mistake (that I made) while tinkering tonight.

It's exactly what it says on the tin: when using LLGL::RenderingDebugger, LLGL will attempt to reflect into the shader. Without `LLGL_D3D12_ENABLE_DXCOMPILER`, a crash occurs.

Crash was due to a lack of graceful HRESULT handling.


LukasBanana:
Looks good to me. Thanks for fixing this.

PR #119, opened 2024-06-14
Fix compilation on Windows

st0rmbtw:
Fix compilation on Windows with a compiler that uses UNIX style arguments (with the `-` prefix instead of `/`)


CMakeLists.txt
```diff
@@ -360,10 +360,13 @@ if(WIN32)
...
+endif(WIN32)
+
+if(MSVC)
     set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} /DLLGL_DEBUG")
 else()
     set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -DLLGL_DEBUG")
-endif(WIN32)
+endif()
 
 ADD_DEBUG_DEFINE(LLGL_DEBUG)
```
LukasBanana, line 371: There is this line which already adds `LLGL_DEBUG` define under the `MSVC` condition (see `ADD_DEBUG_DEFINE` macro). So you can basically just delete the two statements above that modify `CMAKE_CXX_FLAGS_DEBUG`.
Having said that, I read these variables actually shouldn't be touched once the CMake build is running [1].

Can you share the command line or Makefile or whatever build system you use that contains the wrong argument? Or is there any error message causing this line?
I did notice that the `LLGL_DEBUG` define doesn't show up at all in my MSYS Android build, but your change didn't fix that either.

[1] https://alexreinking.com/blog/how-to-use-cmake-without-the-agonizing-pain-part-2.html

LukasBanana (changes requested):

CMakeLists.txt
st0rmbtw: > Can you share the command line or Makefile or whatever build system you use that contains the wrong argument? Or is there any error message caused by this line?

On Windows I use the `Ninja` build system with the `clang` compiler.
`clang` expects arguments prefixed with `-`.
But this code:
```cmake
if(WIN32)
    ADD_DEFINE(_CRT_SECURE_NO_WARNINGS)
    ADD_DEFINE(_SCL_SECURE_NO_WARNINGS)
    ADD_DEFINE(UNICODE)
    ADD_DEFINE(NOMINMAX)
    ADD_DEFINE(_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR)
    if(MSVC)
        # Disable some warnings for MSVC compiler
        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4250 /wd4251 /wd4290 /wd4103")
    endif(MSVC)
    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} /DLLGL_DEBUG")
else()
    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -DLLGL_DEBUG")
endif(WIN32)
```
Adds the `/DLLGL_DEBUG` flag if you're on Windows regardless of the compiler.

So when you compile the project you will get this error:
```
clang++: error: no such file or directory: '/DLLGL_DEBUG'
```

LukasBanana: Thanks for your contribution. Merged with fb2883e

Issue #127, opened 2024-08-14
Metal Build ID mismatch in render system module


st0rmbtw:
`RenderSystem::Load` returns null and `Report` contains `build ID mismatch in render system module`.

That happens because this check fails:
https://github.com/LukasBanana/LLGL/blob/17769f0280ba575745213d0183c1f8dc04cc13e6/sources/Renderer/RenderSystem.cpp#L131

`RenderSystem_BuildID()` returns `210011`, but `LLGL_BUILD_ID` is `200011`.

Looks like `Metal` renderer compiles without `LLGL_DEBUG` define.

PR #128, opened 2024-08-14
Fix OBJCXX debug flags

st0rmbtw:
Metal renderer compiles without the `LLGL_DEBUG` define, because it's written in Objective-C++, but the `ADD_DEBUG_DEFINE` cmake function sets a define only for C++ via `CMAKE_CXX_FLAGS_DEBUG`

Solution:
- Replace modifying `CMAKE_CXX_FLAGS_DEBUG` with the universal `add_compile_definitions` function.

Closes https://github.com/LukasBanana/LLGL/issues/127


LukasBanana:
I keep reading online that modifying `CMAKE_CXX_FLAGS_DEBUG` (and most other CMake specific variables) inside _CMakeLists.txt_ is not intended and can lead to undefined behavior. I can confirm that `LLGL_DEBUG` is not set in Xcode, neither for LLGL_Metal (ObjC++) nor LLGL_OpenGL (C++); it works with VisualStudio but not with Xcode.

It looks as if `add_compile_definitions("$<$<CONFIG:Debug>:${IDENT}>")` is a viable replacement for the macro `ADD_DEBUG_DEFINE`, I would have to increase the minimum CMake version to 3.12, though.

I also wanted to refactor a lot of macro use cases from `#ifdef` to `#if`, which might allow to define the macro for older CMake versions as `add_definitions("-DLLGL_DEBUG=!NDEBUG")` or something similar.

As for your PR, I don't see `CMAKE_OBJCXX_FLAGS_DEBUG` showing up in my Xcode settings, which might be because I have an older version (11.0), but I also never ran into this issue. If your Metal backend compiles without `LLGL_DEBUG` defined, so should the main module `LLGL`. Unless this is one of those undefined behaviors. Since I can't reproduce this issue, can you please try to replace the macro like this:
```cmake
macro(ADD_DEBUG_DEFINE IDENT)
    add_compile_definitions("$<$<CONFIG:Debug>:${IDENT}>")
endmacro()
```

st0rmbtw:
`add_compile_definitions` also works as expected.
Should I update the PR?

LukasBanana:
Before I merge this, I would like to understand how this issue came to be on your Mac, since I can't reproduce it. As mentioned in my previous post, if `LLGL_DEBUG` is not defined in your _LLGL_MetalD.dylib_, it also shouldn't be defined in _LLGLD.dylib_, so this shouldn't have a mismatch. Aren't you building both shared libraries in the same Xcode solution?

Can you please confirm that _MTModuleInterface.mm_ does _not_ have `LLGL_DEBUG` defined while _RenderSystem.cpp_ does have it defined when you build both libraries with **Debug** configuration? Maybe with something like this:
```cpp
// MTModuleInterface.mm
#ifndef LLGL_DEBUG
#error LLGL_DEBUG missing
#endif

// RenderSystem.cpp
#ifndef LLGL_DEBUG
#error LLGL_DEBUG missing
#endif
```

st0rmbtw:
Just checked that, yep, without changes made in this PR it fails in `MTModuleInterface.mm` with `#error LLGL_DEBUG missing`

PR #147, opened 2025-01-13
[Misc] Mark move operators and move constructors noexcept

st0rmbtw:


include/LLGL/Container/SmallVector.h
```diff
@@ -593,7 +593,7 @@ class LLGL_EXPORT SmallVector
             return *this;
         }
 
-        SmallVector& operator = (SmallVector&& rhs)
+        SmallVector& operator = (SmallVector&& rhs) noexcept
```
LukasBanana, line 596: With `SmallVector` and some other classes that depend on it, I was unsure how `noexcept` can be applied here since these operators do a bit more than `std::vector` move operator. There are arbitrary constructors and destructors being invoked. When I was looking for a comparison, I noticed that SPIRV-Cross does the same, though, they also use `noexcept` on their SmallVector implementation: [spirv_cross_containers.hpp:239](https://github.com/KhronosGroup/SPIRV-Cross/blob/main/spirv_cross_containers.hpp#L239)
I'm just wondering how this can be justified considering that any element's copy-constructor could potentially throw an exception.

st0rmbtw: We can decide whether the move assign operator of `SmallVector` is throwable or not by looking at the move constructor of the type of element using `std::is_nothrow_move_constructible<T>::value`.

sources/Core/Image.cpp
```diff
@@ -70,7 +70,7 @@ Image& Image::operator = (const Image& rhs)
     return *this;
 }
 
-Image& Image::operator = (Image&& rhs)
+Image& Image::operator = (Image&& rhs) noexcept
 {
     Reset(rhs.GetExtent(), rhs.GetFormat(), rhs.GetDataType(), std::move(rhs.data_));
```
LukasBanana, line 75: C++ doesn't provide compile-time checks for `noexcept` functions, but as I understand it, any functions being called inside a `noexcept`-declared function _should_ also guarantee to never throw an exception. So that means `Reset()`, `ResetAttributes()`, `GetFormat()`, and `GetDataType()` should also have `noexcept`, which is fine for all other than `Reset()` IMHO. Functions that call `delete` don't look like good `noexcept`-candidates to me, but maybe I'm missing something. I can find arguments for and [against it](https://news.ycombinator.com/item?id=41163083).

LukasBanana (commented):
I'm unsure how reasonable it is to also have all those non-trivial move operators declared with `noexcept`. Have you run some CPU performance analysis that show the benefit of this? I saw you used a thirdparty library for performance captures in your project if I'm not mistaken.

st0rmbtw:
> Have you run some CPU performance analysis that show the benefit of this?

No, I haven't.

I just saw `clangd` giving a warning about move operators not marked `noexcept` and I fixed it.
I can remove `noexcept` from functions you commented above.

include/LLGL/Container/DynamicArray.h
```diff
@@ -360,7 +360,7 @@ class LLGL_EXPORT DynamicArray
             return *this;
         }
 
-        DynamicArray& operator = (DynamicArray&& rhs)
+        DynamicArray& operator = (DynamicArray&& rhs) noexcept
```
LukasBanana, line 363: I think `DynamicArray` isn't a candidate for this either, because it can be used for any kind of elements. While it's not good practice for any class to throw exceptions in the destructor, we would have add a condition into `noexcept()` depending on whether its element type could potentially throw. The other classes look fine, but I think we should skip this one as well.

LukasBanana:
Thanks for bearing with me and thanks again for the contribution.

PR #148, opened 2025-01-13
[Misc] Use move semantics where possible

st0rmbtw:


include/LLGL/FragmentAttribute.h
```diff
@@ -35,23 +35,23 @@ struct FragmentAttribute
...
     {
     }
 
     //! Constructor to initialize all members.
     inline FragmentAttribute(
-        const StringLiteral&    name,
-        const Format            format,
-        std::uint32_t           location    = 0,
-        const SystemValue       systemValue = SystemValue::Undefined)
+        StringLiteral        name,
```
LukasBanana, line 46: What are the performance benefits here? The constructor's parameter will now invoke a copy constructor, will it not? So the copy constructor of `StringLiteral` is either invoked at the caller side or at the callee side.

`StringLiteral` is also very lightweight, as long as the internal string was not dynamically allocated, e.g. when you use `StringLiteral s = std::string("A") + "B"` instead of the lightweight version `StringLiteral s = "AB"`.

sources/Renderer/VertexAttribute.cpp
```diff
@@ -5,6 +5,7 @@
  * Licensed under the terms of the BSD 3-Clause license (see LICENSE.txt).
  */
 
+#include <utility>
```
LukasBanana, line 8: Wouldn't this header also have to be included in _PipelineLayoutFlags.h_? It seems to compile fine, since all CIS builds succeeded, but the correct way would probably be to add this in all other headers that add `std::move`.

sources/Renderer/VertexAttribute.cpp
```diff
@@ -29,51 +30,51 @@ VertexAttribute::VertexAttribute(
...
+    std::uint32_t   location,
+    std::uint32_t   instanceDivisor)
 :
-    name            { semanticName    },
-    format          { format          },
-    location        { location        },
-    semanticIndex   { semanticIndex   },
-    instanceDivisor { instanceDivisor }
+    name            { std::move(semanticName) },
+    format          { format                  },
```
LukasBanana, line 40: I appreciate you following the same formatting style ❤️ I know it may seem a bit nitpicky.

include/LLGL/FragmentAttribute.h
st0rmbtw: > The constructor's parameter will now invoke a copy constructor, will it not?

No, it won't. I know it looks confusing, but it is what it is. There is an [explanation](https://stackoverflow.com/a/51706522) for that on stackoverflow.

sources/Renderer/VertexAttribute.cpp
st0rmbtw: Yeah, forgot to add that one

sources/Renderer/VertexAttribute.cpp
st0rmbtw: No problem :)

include/LLGL/FragmentAttribute.h
LukasBanana: Thanks for the explanation, that's indeed a great improvement. I've been stuck to the pass-by-reference-to-const method for the longest time. I also found another great article explaining this scenario: https://xania.org/202101/cpp-by-value-args
Would you mind also adding `<utility>` to the include directives in _FragmentAttribute.h_?

st0rmbtw: > Would you mind also adding `<utility>` to the include directives in _FragmentAttribute.h_?

Yes, sure.

LukasBanana:
Thanks for your contribution. The last build check only failed because I closed the PR before it could finish.

PR #149, opened 2025-01-19
[GL] Use StringView instead of std::string

st0rmbtw:
This PR is not much useful, just some code quality changes.

I'm not sure if I've done it right, but in theory it should reduce the amount of string copies.

### DIFF
--- a/sources/Renderer/OpenGL/Shader/GLLegacyShader.cpp
+++ b/sources/Renderer/OpenGL/Shader/GLLegacyShader.cpp
@@ -75,7 +75,7 @@ std::string GLLegacyShader::GetGLShaderLog(GLuint shader)
         glGetShaderInfoLog(shader, infoLogLength, &charsWritten, infoLog.data());
 
         /* Convert byte buffer to string */
-        return std::string(infoLog.data(), charsWritten);
+        return std::string(infoLog.data(), static_cast<std::size_t>(charsWritten));
     }
 
     return "";
--- a/sources/Renderer/OpenGL/Shader/GLLegacyShader.cpp
+++ b/sources/Renderer/OpenGL/Shader/GLLegacyShader.cpp
@@ -75,7 +75,7 @@ std::string GLLegacyShader::GetGLShaderLog(GLuint shader)
         glGetShaderInfoLog(shader, infoLogLength, &charsWritten, infoLog.data());
 
         /* Convert byte buffer to string */
-        return std::string(infoLog.data());
+        return std::string(infoLog.data(), charsWritten);
     }
 
     return "";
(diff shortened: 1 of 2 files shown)


sources/Renderer/OpenGL/Shader/GLLegacyShader.cpp
```diff
@@ -75,7 +75,7 @@ std::string GLLegacyShader::GetGLShaderLog(GLuint shader)
         glGetShaderInfoLog(shader, infoLogLength, &charsWritten, infoLog.data());
 
         /* Convert byte buffer to string */
-        return std::string(infoLog.data());
+        return std::string(infoLog.data(), charsWritten);
```
LukasBanana, line 78: Apparently the warning level of all target compilers don't mind this implicit conversation, but I would nonetheless prefer an explicit type cast from `GLsizei` (which is a signed integer) to `std::size_t` here.

sources/Renderer/OpenGL/Shader/GLShaderProgram.cpp
```diff
@@ -168,7 +168,7 @@ std::string GLShaderProgram::GetGLProgramLog(GLuint program)
         glGetProgramInfoLog(program, infoLogLength, &charsWritten, infoLog.data());
 
         /* Convert byte buffer to string */
-        return std::string(infoLog.data());
+        return std::string(infoLog.data(), charsWritten);
```
LukasBanana, line 171: Same here, please wrap this in an explicit type cast.

LukasBanana (changes requested):
Looks like a neat improvement. Once `StringLiteral` was introduced in the public interface, all those `std::string` instantiations become merely transition variables that pass on their internal data but don't need to own the memory, so this change makes perfectly sense.

Please be so good and wrap all the `nameLength` into `static_cast<std::size_t>` to make the type conversion from `GLint` and `GLsizei` explicit. I was suprised the compilers don't warn, but I presume that's only happening on a higher warning level.

sources/Renderer/OpenGL/Shader/GLShaderProgram.cpp
```diff
@@ -482,7 +482,7 @@ static void GLQueryVertexAttributes(GLuint program, ShaderReflection& reflection
                 return true;
             if (lhs.location > rhs.location)
                 return false;
-            return (lhs.name < rhs.name);
+            return lhs.name.compare(rhs.name) < 0;
```
LukasBanana, line 485: This change made me add new operators in `<String.h>` with bdcd2ba, so this line won't have to be changed anymore.

sources/Renderer/OpenGL/Shader/GLShaderProgram.cpp
```diff
@@ -482,7 +482,7 @@ static void GLQueryVertexAttributes(GLuint program, ShaderReflection& reflection
                 return true;
             if (lhs.location > rhs.location)
                 return false;
-            return (lhs.name < rhs.name);
+            return lhs.name < rhs.name;
```
LukasBanana, line 485: Parenthesis was removed, but this line doesn't need any change.

st0rmbtw:
I polluted the branch, so I created a new branch in my local fork, cherry-picked those two commits into that branch and forced pushed them into this remote branch.

Please, check if I've done it right.

LukasBanana:
Looks good to me. I'll merge as soon as the unit tests succeed. GitHub has the _Squash and merge_ operaton, so it shouldn't matter how many commits you pushed.

PR #150, opened 2025-01-22
Make setting a stride of the ImageView data possible

st0rmbtw:
I was implementing the chunked lightmap rendering in my project.
The idea is that instead of one huge lightmap texture, I create multiple smaller ones and render them only if they are visible.
But I ran into a problem when I needed to update the lightmap on the intersection of multiple chunks.
Because I had to calculate proper offsets and size for each chunk's texture.

As the size of the texture of a chunk and the size of the lightmap buffer on CPU are different, I had to do the following:
1. Allocate a buffer with the size of the chunk's texture.
2. Copy the specific portion of data from the CPU lightmap to the bufer.
3. Update the GPU texture with the data from the buffer.
4. Deallocate the buffer.

It was quite slow to do all of that. I could have only the third step if there was an option to set custom stride for the source data of the image view.

But then I noticed that [ID3D11DeviceContext::UpdateSubresource](https://learn.microsoft.com/en-us/windows/win32/api/d3d11/nf-d3d11-id3d11devicecontext-updatesubresource) function has a parameter just for what I need, so I made this PR to expose that parameter in the public interface.

### DIFF
--- a/include/LLGL-C/LLGLWrapper.h
+++ b/include/LLGL-C/LLGLWrapper.h
@@ -1385,6 +1385,7 @@ typedef struct LLGLImageView
     LLGLDataType    dataType; /* = LLGLDataTypeUInt8 */
     const void*     data;     /* = NULL */
     size_t          dataSize; /* = 0 */
+    uint32_t        stride;   /* = 0 */
 }
 LLGLImageView;
 
--- a/include/LLGL-C/LLGLWrapper.h
+++ b/include/LLGL-C/LLGLWrapper.h
@@ -11,9 +11,9 @@
 #define LLGL_C99_LLGLWRAPPER_H
 
 
-#include <stdbool.h>
 #include <stddef.h>
 #include <stdint.h>
+#include <stdbool.h>
 #include <LLGL-C/Types.h>
 
 #if __ANDROID__
@@ -1381,11 +1381,11 @@ LLGLFragmentAttribute;
 
 typedef struct LLGLImageView
 {
-    LLGLImageFormat format;   /* = LLGLImageFormatRGBA */
-    LLGLDataType    dataType; /* = LLGLDataTypeUInt8 */
-    const void*     data;     /* = NULL */
-    size_t          dataSize; /* = 0 */
-    uint32_t        stride;   /* = 0 */
+    LLGLImageFormat format;    /* = LLGLImageFormatRGBA */
+    LLGLDataType    dataType;  /* = LLGLDataTypeUInt8 */
+    const void*     data;      /* = NULL */
+    size_t          dataSize;  /* = 0 */
+    uint32_t        rowStride; /* = 0 */
 }
 LLGLImageView;
 
--- a/include/LLGL/ImageFlags.h
+++ b/include/LLGL/ImageFlags.h
@@ -61,6 +61,10 @@ struct ImageView
 
     //! Specifies the size (in bytes) of the image data. This is primarily used for compressed images and serves for robustness.
     std::size_t dataSize    = 0;
+
+    //! Specifies the size of one row of the image data.
+    //! When this is zero then the width of the texture region is used.
+    std::uint32_t stride = 0;
 };
 
 /**
--- a/include/LLGL/ImageFlags.h
+++ b/include/LLGL/ImageFlags.h
@@ -62,8 +62,11 @@ struct ImageView
     //! Specifies the size (in bytes) of the image data. This is primarily used for compressed images and serves for robustness.
     std::size_t dataSize    = 0;
 
-    //! Specifies the size of one row of the image data.
-    //! When this is zero then the width of the texture region is used.
+    /**
+     \brief Specifies the size of one row of the image data.
+     \remarks When this is zero then the width of the texture region is used.
+     \note Only supported with: Direct3D 12, Direct3D 11, Vulkan, OpenGL.
+    */
     std::uint32_t rowStride = 0;
 };
 
--- a/include/LLGL/ImageFlags.h
+++ b/include/LLGL/ImageFlags.h
@@ -64,7 +64,7 @@ struct ImageView
 
     //! Specifies the size of one row of the image data.
     //! When this is zero then the width of the texture region is used.
-    std::uint32_t stride = 0;
+    std::uint32_t rowStride = 0;
 };
 
 /**
--- a/include/LLGL/ImageFlags.h
+++ b/include/LLGL/ImageFlags.h
@@ -66,7 +66,7 @@ struct ImageView
     /**
      \brief Specifies the size of one row of the image data.
      \remarks When this is zero then the width of the texture region is used.
-     \note Only supported with: Direct3D 12, Direct3D 11, Vulkan, OpenGL.
+     \note Only supported with: Vulkan, OpenGL.
     */
     std::uint32_t rowStride = 0;
 };
--- a/include/LLGL/ImageFlags.h
+++ b/include/LLGL/ImageFlags.h
@@ -42,11 +42,12 @@ struct ImageView
     ImageView(const ImageView&) = default;
 
     //! Constructor to initialize all attributes.
-    inline ImageView(ImageFormat format, DataType dataType, const void* data, std::size_t dataSize) :
-        format   { format   },
-        dataType { dataType },
-        data     { data     },
-        dataSize { dataSize }
+    inline ImageView(ImageFormat format, DataType dataType, const void* data, std::size_t dataSize, std::uint32_t rowStride = 0) :
+        format    { format    },
+        dataType  { dataType  },
+        data      { data      },
+        dataSize  { dataSize  },
+        rowStride { rowStride }
     {
     }
 
--- a/sources/Renderer/Direct3D11/Texture/D3D11Texture.cpp
+++ b/sources/Renderer/Direct3D11/Texture/D3D11Texture.cpp
@@ -278,7 +278,7 @@ HRESULT D3D11Texture::UpdateSubresource(
         dstBox.back   - dstBox.front
     };
 
-    const SubresourceCPUMappingLayout dataLayout = CalcSubresourceCPUMappingLayout(format, extent, numArrayLayers, imageView.format, imageView.dataType);
+    const SubresourceCPUMappingLayout dataLayout = CalcSubresourceCPUMappingLayout(format, extent, numArrayLayers, imageView.format, imageView.dataType, imageView.stride);
 
     if (imageView.dataSize < dataLayout.imageSize)
     {
--- a/sources/Renderer/Direct3D11/Texture/D3D11Texture.cpp
+++ b/sources/Renderer/Direct3D11/Texture/D3D11Texture.cpp
@@ -278,7 +278,7 @@ HRESULT D3D11Texture::UpdateSubresource(
         dstBox.back   - dstBox.front
     };
 
-    const SubresourceCPUMappingLayout dataLayout = CalcSubresourceCPUMappingLayout(format, extent, numArrayLayers, imageView.format, imageView.dataType, imageView.stride);
+    const SubresourceCPUMappingLayout dataLayout = CalcSubresourceCPUMappingLayout(format, extent, numArrayLayers, imageView.format, imageView.dataType, imageView.rowStride);
 
     if (imageView.dataSize < dataLayout.imageSize)
     {
--- a/sources/Renderer/Direct3D11/Texture/D3D11Texture.cpp
+++ b/sources/Renderer/Direct3D11/Texture/D3D11Texture.cpp
@@ -278,7 +278,7 @@ HRESULT D3D11Texture::UpdateSubresource(
         dstBox.back   - dstBox.front
     };
 
-    const SubresourceCPUMappingLayout dataLayout = CalcSubresourceCPUMappingLayout(format, extent, numArrayLayers, imageView.format, imageView.dataType, imageView.rowStride);
+    const SubresourceCPUMappingLayout dataLayout = CalcSubresourceCPUMappingLayout(format, extent, numArrayLayers, imageView.format, imageView.dataType);
 
     if (imageView.dataSize < dataLayout.imageSize)
     {
--- a/sources/Renderer/OpenGL/Texture/GLTexSubImage.cpp
+++ b/sources/Renderer/OpenGL/Texture/GLTexSubImage.cpp
@@ -428,8 +428,6 @@ bool GLTexSubImage(
     if (IsCompressedFormat(imageView.format) && !HasExtension(GLExt::ARB_texture_compression))
         return false;
 
-    glPixelStorei(GL_UNPACK_ROW_LENGTH, imageView.rowStride);
-
     switch (type)
     {
         #ifdef LLGL_OPENGL
@@ -469,8 +467,6 @@ bool GLTexSubImage(
         default:
             return false;
     }
-
-    glPixelStorei(GL_UNPACK_ROW_LENGTH, 0);
 
     return true;
 }
--- a/sources/Renderer/OpenGL/Texture/GLTexSubImage.cpp
+++ b/sources/Renderer/OpenGL/Texture/GLTexSubImage.cpp
@@ -428,7 +428,7 @@ bool GLTexSubImage(
     if (IsCompressedFormat(imageView.format) && !HasExtension(GLExt::ARB_texture_compression))
         return false;
 
-    glPixelStorei(GL_UNPACK_ROW_LENGTH, imageView.stride);
+    glPixelStorei(GL_UNPACK_ROW_LENGTH, imageView.rowStride);
 
     switch (type)
     {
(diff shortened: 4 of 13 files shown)


include/LLGL/ImageFlags.h
```diff
@@ -61,6 +61,10 @@ struct ImageView
 
     //! Specifies the size (in bytes) of the image data. This is primarily used for compressed images and serves for robustness.
     std::size_t dataSize    = 0;
+
+    //! Specifies the size of one row of the image data.
+    //! When this is zero then the width of the texture region is used.
+    std::uint32_t stride = 0;
```
LukasBanana, line 67: Could you please rename this to `rowStride` so we can later add `layerStride` which can be translated to `GL_UNPACK_IMAGE_HEIGHT` in GL and D3D has `SrcDepthPitch` for instance. We can add this later, adding `rowStride` is difficult enough for 5 backends :)

Please also add a new parameter for this field to the constructor with a default value of 0.

Lastly, I don't see this implemented for Metal. I can help with that but until then, please change the comment to this:
```cpp
/**
\brief Specifies the size of one row of the image data.
\remarks When this is zero then the width of the texture region is used.
\note Only supported with: Direct3D 12, Direct3D 11, Vulkan, OpenGL.
*/
```

include/LLGL-C/LLGLWrapper.h
```diff
@@ -1385,6 +1385,7 @@ typedef struct LLGLImageView
     LLGLDataType    dataType; /* = LLGLDataTypeUInt8 */
     const void*     data;     /* = NULL */
     size_t          dataSize; /* = 0 */
+    uint32_t        stride;   /* = 0 */
```
LukasBanana, line 1388: I presume you tripped the static assertion that ensures that the C++ interface remains compatible with the C99 wrapper. Adding this manual is not adviced (see comment at the top of this file) but generally okay for such a small change. Having said that, the C# and Go wrappers would also need this new field and especially the C# wrapper is a bit more complicated.

My current workflow to update the language bindings for new public interfaces is to run `scripts/GenerateWrappers.bat` Windows batch script and copy the emitted _LLGLWrapper.*_ files into their respective folders:
* _includes/LLGL-C/LLGLWrapper.h_
* _wrapper/CSharp/LLGLWrapper.cs_
* _wrapper/Go/LLGLWrapper.go_

You'll need Python 3 installed to run the script.

sources/Renderer/TextureUtils.h
```diff
@@ -64,17 +64,19 @@ LLGL_EXPORT SubresourceCPUMappingLayout CalcSubresourceCPUMappingLayout(
     const Extent3D&     extent,
     std::uint32_t       numArrayLayers,
     const ImageFormat   imageFormat,
-    const DataType      imageDataType
+    const DataType      imageDataType,
+    uint32_t            srcDataStride = 0
```
LukasBanana, line 68: I once thought about changing all the `std::uint32_t` types back to just plain C `uint32_t` but I originally wanted the code base to be explicitly C++ and now I'm stuck with it. So I'd prefer to stick with the one that explicitly refers to C++'s `std` namespace.

sources/Renderer/TextureUtils.cpp
```diff
@@ -72,12 +72,13 @@ static void CalcSubresourceLayoutPrimary(
...
     const Extent3D&     extent,
-    std::uint32_t       numArrayLayers)
+    std::uint32_t       numArrayLayers,
+    std::uint32_t       srcDataStride = 0)
 {
     const FormatAttributes& formatAttribs = GetFormatAttribs(format);
     if (formatAttribs.blockWidth > 0 && formatAttribs.blockHeight > 0)
     {
-        outLayout.rowStride         = (extent.width * formatAttribs.bitSize) / formatAttribs.blockWidth / 8;
+        outLayout.rowStride         = ((srcDataStride > 0 ? srcDataStride : extent.width) * formatAttribs.bitSize) / formatAttribs.blockWidth / 8;
```
LukasBanana, line 81: How should LLGL behave when the row stride is smaller than the width of the source image? Should this be undefined behavior or be defined as an invalid argument? Either way, it should be documented in the `ImageView` struct for this field. If we consider it an invalid argument we should probably also add `max(srcDataStride, extent.width)` here to clamp it to the largest value or add an assertion `LLGL_ASSERT(srcDataStride == 0 || srcDataStride >= extent.width)`.

sources/Renderer/Vulkan/VKRenderSystem.cpp
```diff
@@ -961,6 +961,45 @@ VKDeviceBuffer VKRenderSystem::CreateStagingBufferAndInitialize(
...
+    const Extent3D&             extent)
+{
+    /* Allocate staging buffer */
+    VKDeviceBuffer stagingBuffer = CreateStagingBuffer(createInfo);
+
+    const void* data = srcImageView.data;
+    const std::size_t dataSize = srcImageView.dataSize;
+
+    /* Copy initial data to buffer memory */
+    if (data != nullptr && dataSize > 0) {
```
LukasBanana, line 976: Please use [Allman](https://en.wikipedia.org/wiki/Indentation_style#Notable_styles) indentation in this project instead of K&R.

sources/Renderer/Vulkan/VKRenderSystem.cpp
```diff
@@ -961,6 +961,45 @@ VKDeviceBuffer VKRenderSystem::CreateStagingBufferAndInitialize(
...
+            /* Map buffer memory to host memory */
+            VKDeviceMemory* deviceMemory = region->GetParentChunk();
+            if (void* memory = deviceMemory->Map(device_, region->GetOffset(), dataSize))
+            {
+                if (srcImageView.stride > 0) {
+                    const std::uint32_t stride = GetMemoryFootprint(srcImageView.format, srcImageView.dataType, srcImageView.stride);
+                    const std::size_t width = GetMemoryFootprint(srcImageView.format, srcImageView.dataType, extent.width);
+
+                    for (std::uint32_t row = 0; row < extent.height; ++row) {
+                        ::memcpy(static_cast<std::uint8_t*>(memory) + row * width, static_cast<const std::uint8_t*>(data) + row * stride, width);
```
LukasBanana, line 988: You should be able to replace the `memcpy`s and this for-loop with `BitBlit()`, it's defined in _sources/Core/ImageUtils.h_ and can handle row and layer strides.

sources/Renderer/OpenGL/Texture/GLTexSubImage.cpp
```diff
@@ -428,6 +428,8 @@ bool GLTexSubImage(
     if (IsCompressedFormat(imageView.format) && !HasExtension(GLExt::ARB_texture_compression))
         return false;
 
+    glPixelStorei(GL_UNPACK_ROW_LENGTH, imageView.stride);
```
LukasBanana, line 431: The GL pixel store parameters are managed by `GLStateManager` (see `GLStateManager::SetPixelStoreUnpack`) and _should_ therefore not be changed without it. Even though you set it back to 0, this might not be the value the state manager memorized it as. You could simply use `GLStateManager::Get().SetPixelStoreUnpack()` to access the active state manager, but I don't think it's a good idea to pull the state manager into these functions unless we refactor some other uses of it. If you look at `GLTexture::CopyImageFromBuffer()` you'll see that it already uses said function to define the row stride for a buffer-to-texture operation and then immediately calls either `GLTextureSubImage()` or `GLTexSubImage()`, so this might break.

This functionality is also missing from `GLTexImage` and `GLTextureImage`, the ones that allocate the GL texture storage. With a new attribute called `stride` (or `rowStride`) I would expect this to work with both `CreateTexture()` and `WriteTexture()`, but GL would only support the latter one.

With all that being said, I think it's best we move this parameter out of this function and put it at all the locations where it's needed; shouldn't be too many. Actually, I only see `GLTexture::AllocTextureStorage()` and `GLTexture::TextureSubImage()` as the two places and that should make it pretty easy.

LukasBanana (changes requested):
I'm glad to see you getting acquainted with the inner workings of LLGL :-) I started with a concept for `rowStride` and `layerStride` in this struct earlier of last year but never took the time to start on the implementation because I had no need for this feature so far.

I'm happy to help with the Metal backend, but we can add this a bit later. It would make sense to also add this field to `MutableImageView` at some point, but that's maybe something for another time.

sources/Renderer/Vulkan/VKRenderSystem.cpp
```diff
@@ -477,7 +477,7 @@ void VKRenderSystem::WriteTexture(Texture& texture, const TextureRegion& texture
         VK_BUFFER_USAGE_TRANSFER_SRC_BIT // <-- TODO: support read/write mapping //GetStagingVkBufferUsageFlags(bufferDesc.cpuAccessFlags)
     );
 
-    VKDeviceBuffer stagingBuffer = CreateStagingBufferAndInitialize(stagingCreateInfo, imageData, imageDataSize);
+    VKDeviceBuffer stagingBuffer = CreateTextureStagingBufferAndInitialize(stagingCreateInfo, srcImageView, extent);
```
LukasBanana, line 480: It looks like this change ignores `imageData`. This function must not use the input parameter directly in case the image format and data type had to be converted (see `intermediateData`). That's what all the code above is for. Oh, and the convert functions will also have to handle the new stride. Yeah, this features goes wide I'm afraid 😬

st0rmbtw:
It was a quick and a low effort solution just to get it working, and also I was sleepy 😅. I just wanted to get your feedback on the idea and implementation.

> I'm glad to see you getting acquainted with the inner workings of LLGL :-)

Yeah, now I understand it a little better :)

include/LLGL-C/LLGLWrapper.h
st0rmbtw: > I presume you tripped the static assertion that ensures that the C++ interface remains compatible with the C99 wrapper.

Yeah, you're right 😄. I forgot that there are more language bindings than just C. And also didn't know such a tool existed.
Now I know :)

sources/Renderer/TextureUtils.cpp
st0rmbtw: I'm not sure why it is an issue. Isn't the extent there describes the size of the **destination** subresource? If so then it doesn't matter bigger or smaller the row stride is. Correct me if I'm wrong, please.

sources/Renderer/Vulkan/VKRenderSystem.cpp
st0rmbtw: I fixed the ignoring of `imageData`, but I'm not sure about `ConvertImageBuffer` because I can't figure out how to take the stride into account in `DoConcurrentRange`

LukasBanana:
I'll take another look over the weekend since this is a bit more involved and I'll come up with a unit test for this feature.

sources/Renderer/Vulkan/VKRenderSystem.cpp
LukasBanana: I'll see to also update the image conversion functions over the weekend. The work distribution likely needs to be adjusted.

sources/Renderer/TextureUtils.cpp
LukasBanana: If the row stride is smaller than the extent, which is used for both source and destination subresource, copy operations will have overlapping ranges of memory they are trying to copy to. This will lead to undefined behavior when used in parallel either with the multi-threaded image conversion functions or GPU operations.
I am just wondering if we should have an assertion here (e.g. `stride == 0 || stride >= width`) or just clamp it via `max(stride, width)`.

`CommandBuffer::CopyBufferFromTexture()` for instance has this contract with its `rowStride` parameter:
```
If \c rowStride is 0, the source data is considered to be tightly packed for each array layer and the required alignment is managed automatically.
If \c rowStride is not 0, it \b must be greater than or equal to the size (in bytes) of each row in the texture region with respect to the texture's format.
```

I might have forgotten to put any safe guards in the code for this parameter, but it should at least be documented what the expected behavior is for these cases. I think we should consider a stride smaller than the width of a copy operation to be an invalid parameter and therefore put in an assertion. This way, the programmer will be confronted with this issue right away.

sources/Renderer/Vulkan/VKRenderSystem.cpp
```diff
@@ -961,6 +963,39 @@ VKDeviceBuffer VKRenderSystem::CreateStagingBufferAndInitialize(
...
+    /* Copy initial data to buffer memory */
+    if (data != nullptr && dataSize > 0)
+    {
+        if (VKDeviceMemoryRegion* region = stagingBuffer.GetMemoryRegion())
+        {
+            /* Map buffer memory to host memory */
+            VKDeviceMemory* deviceMemory = region->GetParentChunk();
+            if (void* memory = deviceMemory->Map(device_, region->GetOffset(), dataSize))
+            {
+                const char* src = static_cast<const char*>(data);
```
LukasBanana, line 986: I guess I learned something new here :-)  I was using `reinterpret_cast` for these types of conversions, but this `static_cast` seems just fine and is also the better choice (found [this on Stackoverflow](https://stackoverflow.com/questions/310451/should-i-use-static-cast-or-reinterpret-cast-when-casting-a-void-to-whatever)).

LukasBanana (changes requested):
I added just one small change request. Besides that, I think this is good to go. I will fiddle with the image conversion functions afterwards. I hope you don't mind me pulling you into a new discussion, though, if some future tests fail because of this implementation 😄

sources/Renderer/Direct3D11/Texture/D3D11Texture.cpp
```diff
@@ -278,7 +278,7 @@ HRESULT D3D11Texture::UpdateSubresource(
         dstBox.back   - dstBox.front
     };
 
-    const SubresourceCPUMappingLayout dataLayout = CalcSubresourceCPUMappingLayout(format, extent, numArrayLayers, imageView.format, imageView.dataType);
+    const SubresourceCPUMappingLayout dataLayout = CalcSubresourceCPUMappingLayout(format, extent, numArrayLayers, imageView.format, imageView.dataType, imageView.rowStride);
```
LukasBanana, line 281: This won't work until `ConvertImageBuffer()` below implements this parameter, too. So I suggest we skip D3D11 and D3D12 for this PR. I'll implement this afterwards and then we can add those backends, too. Please also update the documentation of what backends will be supported initially in the comment for `ImageView::rowStride`.

sources/Renderer/TextureUtils.cpp
st0rmbtw: > I am just wondering if we should have an assertion here (e.g. `stride == 0 || stride >= width`) or just clamp it via `max(stride, width)`.

I think an assertion would be better (maybe even put it in `DbgRenderSystem::WriteTexture`?)

sources/Renderer/Direct3D11/Texture/D3D11Texture.cpp
st0rmbtw: Ohh, okay :(

sources/Renderer/Vulkan/VKRenderSystem.cpp
st0rmbtw: I can make a PR to replace `reinterpret_cast` with `static_cast` when converting from `void*` :)

sources/Renderer/Direct3D11/Texture/D3D11Texture.cpp
LukasBanana: Those changes weren't that big and we can get them in next in a separate CL, but this would be half-implemented only, so I rather have it properly supported there or not (for now).

sources/Renderer/Vulkan/VKRenderSystem.cpp
```diff
@@ -477,7 +479,7 @@ void VKRenderSystem::WriteTexture(Texture& texture, const TextureRegion& texture
         VK_BUFFER_USAGE_TRANSFER_SRC_BIT // <-- TODO: support read/write mapping //GetStagingVkBufferUsageFlags(bufferDesc.cpuAccessFlags)
     );
 
-    VKDeviceBuffer stagingBuffer = CreateStagingBufferAndInitialize(stagingCreateInfo, imageData, imageDataSize);
+    VKDeviceBuffer stagingBuffer = CreateTextureStagingBufferAndInitialize(stagingCreateInfo, extent, imageData, imageDataSize, srcImageView.rowStride, bytesPerPixel);
```
LukasBanana, line 482: `GetMemoryFootprint()` will return 0 if `numTexels` is smaller than the block size, so `bytesPerPixel` will be 0 in this code path for block compression formats such as `BC1UNorm` (also see [`FormatAttributes::blockWidth`](https://github.com/LukasBanana/LLGL/blob/master/include/LLGL/Format.h#L437)).

For now, we should keep the old function when a compressed format is specified, because it doesn't rely on `bytesPerPixel`. You can use `IsCompressedFormat(format)` for this.

LukasBanana (changes requested):
Sorry I missed this in my last review, but we need to handle the case for compressed formats, because they are all block-compressions, i.e. the value of "bytes per pixel" makes no sense in that context since they are never sampled for a single pixel. For those functions, we should skip `rowStride` and use the old function to just copy the whole input data.

sources/Renderer/Vulkan/VKRenderSystem.cpp
st0rmbtw: Oh, okay

LukasBanana:
Thank you, I'll submit as soon as the CI builds go through.

PR #151, opened 2025-01-23
Create macros for throwing to avoid boilerplate code

st0rmbtw:
This PR adds a new `Throw.h` file with the `LLGL_THROW` and `LLGL_THROW_IF` macros that work only if the `LLGL_DEBUG` or the `LLGL_ENABLE_EXCEPTIONS` are defined. So now LLGL can build with the `-fno-exceptions` compiler flag.


sources/Platform/Linux/LinuxDisplay.cpp
```diff
@@ -57,8 +58,7 @@ static void* g_retainedLibGL = nullptr;
 LinuxSharedX11Display::LinuxSharedX11Display() :
     native_ { XOpenDisplay(nullptr) }
 {
-    if (!native_)
-        throw std::runtime_error("failed to open connection to X server");
+    LLGL_THROW_IF(!native_, std::runtime_error("failed to open connection to X server"));
```
LukasBanana, line 61: For any remaining exceptions inside source files under `sources/` I'd prefer to use `LLGL_TRAP` (see _sources/Core/Exception.h_) or `LLGL_ASSERT` (see _sources/Core/Assertion.h_). It throws an `std::runtime_error` when exceptions are enabled, breaks the debugger, or calls `abort()` and provides a formatted callstack.

sources/Renderer/Metal/RenderState/MTGraphicsPSO.mm
```diff
@@ -192,14 +193,17 @@ static void FillColorAttachmentDesc(
     if (hasTessellationStage)
     {
         vertexShaderMT = LLGL_CAST(const MTShader*, desc.tessEvaluationShader);
-        if (vertexShaderMT == nullptr || !vertexShaderMT->IsPostTessellationVertex())
-            throw std::invalid_argument("cannot create Metal tessellation pipeline without post-tessellation vertex shader (in 'tessEvaluationShader')");
+
+        LLGL_THROW_IF(
```
LukasBanana, line 197: An exception like this under the _sources/_ folder can be rewritten like so:
```cpp
LLGL_ASSERT(
    vertexShaderMT != nullptr && vertexShaderMT->IsPostTessellationVertex(),
    "cannot create Metal tessellation pipeline without post-tessellation vertex shader (in 'tessEvaluationShader')"
);
```

LukasBanana (changes requested):
I appreciate the cleanup work here. Another jobsite I wanted to pickup a long time ago and only did so incrementally. I'm fine with the new macro in the public interface for the few places with exceptions, like the container classes, but for all internal code, I'd prefer to stick with the `LLGL_TRAP`/`LLGL_ASSERT` macros as described in the comments.

include/LLGL/Throw.h
```diff
@@ -0,0 +1,31 @@
...
+#define LLGL_THROW_H
+
+#if __EXCEPTIONS || __cpp_exceptions == 199711
+    #define LLGL_COMPILER_EXCEPTIONS_ENABLED 1
+#endif
+
+#if (LLGL_DEBUG || LLGL_ENABLE_EXCEPTIONS) && LLGL_COMPILER_EXCEPTIONS_ENABLED
+    #define LLGL_THROW(EXCEPTION) throw (EXCEPTION)
+    #define LLGL_THROW_IF(CONDITION, EXCEPTION) if (CONDITION) { LLGL_THROW(EXCEPTION); }
+    #define LLGL_VERIFY_OR_THROW(CONDITION, EXCEPTION) LLGL_THROW_IF(!(CONDITION), EXCEPTION)
```
LukasBanana, line 18: I would go only with `LLGL_VERIFY_OR_THROW`. I meant to replace all use cases with the opposite condition, so all of them act like an assertion that their condition must be met or otherwise an exception is thrown.

With `LLGL_THROW()` replaced by `LLGL_TRAP()` and `LLGL_ASSERT()` inside the _sources/_ folder, we can boil this down to only one new macro in the public interface.

sources/Renderer/Metal/RenderState/MTGraphicsPSO.mm
```diff
@@ -192,14 +193,17 @@ static void FillColorAttachmentDesc(
...
+            std::invalid_argument("cannot create Metal tessellation pipeline without post-tessellation vertex shader (in 'tessEvaluationShader')")
+        );
     }
     else
     {
         vertexShaderMT = LLGL_CAST(const MTShader*, desc.vertexShader);
-        if (vertexShaderMT == nullptr)
-            throw std::invalid_argument("cannot create Metal pipeline without vertex shader");
+
+        LLGL_THROW_IF(vertexShaderMT == nullptr, std::invalid_argument("cannot create Metal pipeline without vertex shader"));
```
LukasBanana, line 206: There are still several uses of this new macro in the `sources/` folder. I'm okay with replacing the `std::invalid_argument` exception with `LLGL_TRAP()` or `LLGL_ASSERT()`. My rationale here is that LLGL is supposed to use very little exceptions anyway and the little number of exceptions that are thrown can be unified with a thin abstraction to provide helpful callstacks when exceptions are disabled.

include/LLGL/TypeInfo.h
```diff
@@ -97,8 +98,8 @@ const LLGL::Window& myWindow = LLGL::CastTo<LLGL::Window>(mySurface);
 template <typename T>
 inline const T& CastTo(const Interface& obj)
 {
-    if (!IsInstanceOf<T>(obj))
-        throw std::bad_cast();
+    LLGL_THROW_IF(!IsInstanceOf<T>(obj), std::bad_cast());
```
LukasBanana, line 101: Was this left unchanged on purpose? The other version with `LLGL_VERIFY_OR_THROW` looks good to me.

sources/Renderer/Metal/MTCore.mm
```diff
@@ -23,8 +24,8 @@ void MTThrowIfFailed(NSError* error, const char* info)
         
         NSString* errorMsg = [error localizedDescription];
         s += [errorMsg cStringUsingEncoding:NSUTF8StringEncoding];
-        
-        throw std::runtime_error(s);
+
+        LLGL_TRAP(s.c_str());
```
LukasBanana, line 28: This is a pitfall because `LLGL_TRAP()` expects a format string just like `printf()`, so unless you can guarantee that the error message from `NSError` never contains the `%` character this can crash because no corresponding argument is provided in the variadic argument list. So this needs to be `LLGL_TRAP("%s", s.c_str())`.

LukasBanana (changes requested):
I think we can boil this down to just one new macro in the public interface. I left some more comments in the code changes.

include/LLGL/Throw.h
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+#ifndef LLGL_THROW_H
+#define LLGL_THROW_H
+
+#if __EXCEPTIONS || __cpp_exceptions == 199711
+    #define LLGL_COMPILER_EXCEPTIONS_ENABLED 1
+#endif
+
+#if (LLGL_DEBUG || LLGL_ENABLE_EXCEPTIONS) && LLGL_COMPILER_EXCEPTIONS_ENABLED
```
LukasBanana, line 15: I still don't think `LLGL_DEBUG` makes sense here. You might want a debug build but turn off exceptions entirely.

include/LLGL/TypeInfo.h
st0rmbtw: Oops

sources/Renderer/OpenGL/Platform/MacOS/MacOSGLContext.mm
```diff
@@ -48,8 +50,7 @@
     Surface&                            surface,
     MacOSGLContext*                     sharedContext)
 {
-    if (!CreatePixelFormat(pixelFormat, profile))
-        throw std::runtime_error("failed to find suitable OpenGL pixel format");
+    LLGL_ASSERT(CreatePixelFormat(pixelFormat, profile), "failed to find suitable OpenGL pixel format");
```
LukasBanana, line 53: Assertion must not be used to mandatory expressions, just for debugging expressions. Assertions could be optimized out if you were to decide to replace them with a dummy macro `LLGL_ASSERT(...) /*No substitue*/`. But this conditional expression must always be executed. So this needs to be a `LLGL_TRAP()` instead.

LukasBanana (changes requested):
I found one last issue, the rest looks good. Once you updated this last bit, I'll merge tomorrow after the CIS build went through.

sources/Renderer/OpenGL/Platform/MacOS/MacOSGLContext.mm
st0rmbtw: Oh, yeah. It could be a so hard to find bug.

LukasBanana:
I made some changes to the new exception throwing with e8d86d7. I hope you don't feel blindsided, I appreciate you getting the ball rolling on this matter. The behavior hasn't really changed, this will still throw exceptions if enabled, but in the case they are disabled, this provides a nice callstack to the standard error or breaks the debugger if one is attached. The only downside is that this change only accepts a handful of predefined exceptions such as `std::invalid_argument`, `std::bad_cast` etc., but I think that's fine since LLGL does handle any exceptions at all, it only throws them which allows the client programmer to exit the application more gracefully than with an abort.

PR #153, opened 2025-01-26
Implement custom src row stride for the D3D11 and D3D12 backends

st0rmbtw:
## Description
This PR is a workaround until we figure out how to make `ConvertImageBuffer` respect src row stride.
It works. Checked `D3D11`, `D3D12` and `Vulkan` on Windows.

## Solution
If source image data needs to be converted and src row stride is not equal to the texture width, then we do the following:
1. Allocate a new buffer, so then in `ConvertImageBuffer` data is converted correctly.
2. Copy the source data respecting the stride to that buffer.
3. Call `ConvertImageBuffer` on that buffer.

**P.S.:** Please, mention me via `@`, so I can get a notification on my phone that you left a comment and reply to you as soon as possible.

### DIFF
--- a/sources/Core/ImageFlags.cpp
+++ b/sources/Core/ImageFlags.cpp
@@ -211,13 +211,20 @@ static void ConvertImageBufferDataTypeWorker(
     const std::uint32_t numComponents       = ImageFormatSize(srcImageView.format);
     const std::uint32_t numComponentsPerRow = extent.width*numComponents;
 
+    const std::size_t begin = idxBegin * numComponents;
+    const std::size_t end   = idxEnd * numComponents;
+
     VariantConstBuffer  srcBuffer = srcImageView.data;
     VariantBuffer       dstBuffer = dstImageView.data;
 
-    for_subrange(i, idxBegin*numComponents, idxEnd*numComponents)
+    const std::size_t y = idxBegin / extent.width;
+    srcBuffer.int8 += y * srcRowPadding;
+    dstBuffer.int8 += y * dstRowPadding;
+
+    for_subrange(i, begin, end)
     {
         /* Apply source and destination stride when passing an edge */
-        if (i > 0 && i % numComponentsPerRow == 0)
+        if (i > begin && i % numComponentsPerRow == 0)
         {
             srcBuffer.int8 += srcRowPadding;
             dstBuffer.int8 += dstRowPadding;
--- a/sources/Core/ImageFlags.cpp
+++ b/sources/Core/ImageFlags.cpp
@@ -557,20 +557,24 @@ static void ConvertImageBufferFormatWorker(
         SetVariantMinMax(srcImageView.dataType, colorValue.b, true);
         SetVariantMinMax(srcImageView.dataType, colorValue.a, false);
 
-        for_subrange(i, begin, end)
+        for_subrange(k, begin, end)
         {
+            const size_t rowStride = GetImageRowStride(srcImageView, extent);
+            const size_t y = k / extent.width;
+            const size_t i = k % extent.width;
+
             /* Apply source and destination stride when passing an edge */
-            if (i > 0 && i % extent.width == 0)
+            if (k > 0 && i == 0)
             {
-                srcBuffer.int8 += srcRowPadding;
+                // srcBuffer.int8 += srcRowPadding;
                 dstBuffer.int8 += dstRowPadding;
             }
 
             /* Read RGBA variant from source buffer */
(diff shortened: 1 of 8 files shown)


LukasBanana:
I submitted f71d0db which implements `rowStride` for all `ConvertImageBuffer()` functions. @st0rmbtw  Could you please sync and see if this simplifies your addition?

st0rmbtw:
@LukasBanana It doesn't really work :/. Looks like `ConvertImageBuffer` calculates wrong offsets or something.

sources/Core/ImageFlags.cpp
```diff
@@ -534,7 +538,7 @@ static void ConvertImageBufferFormatWorker(
         for_subrange(i, begin, end)
         {
             /* Apply source and destination stride when passing an edge */
-            if (i > 0 && i % extent.width == 0)
+            if (i > begin && i % extent.width == 0)
```
LukasBanana, line 541: Good catch!

sources/Core/ImageFlags.cpp
```diff
@@ -526,6 +526,10 @@ static void ConvertImageBufferFormatWorker(
     VariantConstBuffer  srcBuffer = srcImageView.data;
     VariantBuffer       dstBuffer = dstImageView.data;
 
+    const std::size_t y = begin / extent.width;
```
LukasBanana, line 529: Makes sense to me, I didn't think of the previous padding that would have been necessary when the worker starts half way through the buffer. I knew this will need more testing, but I only came up with a quick and simple unit test so far.

sources/Renderer/Direct3D11/Texture/D3D11Texture.cpp
```diff
@@ -296,12 +297,17 @@ HRESULT D3D11Texture::UpdateSubresource(
     const char* srcData = static_cast<const char*>(imageView.data);
     LLGL_ASSERT_PTR(srcData);
 
+    std::uint32_t srcRowStride      = imageView.rowStride > 0 ? imageView.rowStride                 : dataLayout.rowStride;
```
LukasBanana, line 300: Would there ever be a reason to have `imageView.rowStride` be non-zero but smaller than `dataLayout.rowStride`? If not, maybe the maximum between the two values is a reasonable alternative. Let me know what you think, I don't have a strong opinion on this one.

sources/Renderer/Direct3D12/D3D12RenderSystem.cpp
```diff
@@ -21,6 +21,7 @@
 #include <LLGL/Backend/Direct3D12/NativeHandle.h>
 #include <limits.h>
 
+#include "LLGL/Container/DynamicArray.h"
```
LukasBanana, line 24: This header should be included as system header via `<>`.

sources/Renderer/RenderSystem.cpp
```diff
@@ -26,6 +26,7 @@
 #include <unordered_map>
 
 #include "../Core/PrintfUtils.h"
+#include "TextureUtils.h"
```
LukasBanana, line 29: If I haven't missed anything, I believe this and the two _TextureUtils.*_ files don't need to be changed anymore.

sources/Renderer/Vulkan/VKRenderSystem.cpp
```diff
@@ -445,11 +460,15 @@ void VKRenderSystem::WriteTexture(Texture& texture, const TextureRegion& texture
     /* Check if image data must be converted */
     DynamicByteArray intermediateData;
 
+    std::uint32_t srcRowStride = srcImageView.rowStride > 0 ? srcImageView.rowStride : extent.width * bytesPerPixel;
```
LukasBanana, line 463: Same question here as above. Let me know if you think a `max()` would make sense here.

LukasBanana (changes requested):
Awesome work @st0rmbtw! I left a few comments for some minor change requests.

st0rmbtw:
@LukasBanana Should I update the `ConvertImageBufferDataTypeWorker` function same way as I did in [5a1d390](https://github.com/LukasBanana/LLGL/pull/153/commits/5a1d390269deee9af607b63f6e72a2cd7984d37d)?

LukasBanana:
> Should I update the ConvertImageBufferDataTypeWorker function

Yes, that would likely need the same fix.

st0rmbtw:
> Thanks again for this great contribution ❤️

No problem! I'm happy to help make this library better! I wish more people knew about it.

PR #154, opened 2025-01-28
[GL] Provide rowStride for SetPixelStoreUnpack in pixels instead of bytes

st0rmbtw:
This PR fixes OpenGL crashing because `GL_UNPACK_ROW_LENGTH` expects the number of pixels not bytes.

### DIFF
--- a/sources/Renderer/OpenGL/Texture/GLTexture.cpp
+++ b/sources/Renderer/OpenGL/Texture/GLTexture.cpp
@@ -22,6 +22,7 @@
 #include "../../TextureUtils.h"
 #include "../../../Core/Exception.h"
 #include "../../../Core/CoreUtils.h"
+#include <LLGL/Format.h>
 #include <LLGL/Utils/ForRange.h>
 #include <LLGL/Backend/OpenGL/NativeHandle.h>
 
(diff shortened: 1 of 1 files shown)


sources/Renderer/OpenGL/Texture/GLTexture.cpp
```diff
@@ -866,7 +867,9 @@ void GLTexture::TextureSubImage(const TextureRegion& region, const ImageView& sr
 {
     if (!IsRenderbuffer())
     {
-        GLStateManager::Get().SetPixelStoreUnpack(srcImageView.rowStride, region.extent.height, 1);
+        const std::uint32_t srcRowStride = srcImageView.rowStride / GetMemoryFootprint(srcImageView.format, srcImageView.dataType, 1);
```
LukasBanana, line 870: Please keep in mind that compressed formats are handled differently and are not supported for row strides at the moment, so `GetMemoryFootprint` will return 0 for such formats. You can simply use `0` as a result when the bytes per pixel are also `0`:
```
bytePerPixel > 0 ? rowStride / bytesPerPixel : 0
```

LukasBanana (changes requested):
Changes make sense, but we need to handle the case for compressed formats again.

sources/Renderer/OpenGL/Texture/GLTexture.cpp
st0rmbtw: Oh, right! I keep forgetting about that.

Issue #145, opened 2025-01-11
Rendering glitches with the `Vulkan` and `D3D12` backends


st0rmbtw:
I have a compute shader that writes to a texture.

With the `D3D11` backend, it renders perfectly fine, no glitches at all. But with the either `Vulkan` or `D3D12` backends, there are big artifacts in every frame.

The light rendered with the `D3D11` backend:
![image](https://github.com/user-attachments/assets/32233e11-c0ba-44a0-a2a6-1468c41b6b1d)

The light rendered with the `D3D12` backend (same with `Vulkan`):
![image](https://github.com/user-attachments/assets/ccbf4b2c-33ef-492f-8bc5-424cf06cc79d)

Normal light texture (`D3D11`):
![image](https://github.com/user-attachments/assets/23e03347-9114-4ccb-991b-e70d02abd20b)

Broken light texture (`D3D12`):
![image](https://github.com/user-attachments/assets/0fa055de-9730-4aef-9320-22eed35bb92c)

I've uploaded RenderDoc captures to an online file exchanger:
https://limewire.com/d/bfead59b-00e4-4427-beb7-ffcc0f432270#PpM6H2RwjThP17f9--DMVmVZ2qNgau15qY1-yOgZkAg

LukasBanana:
This is a resource barrier issue which is currently only handled correctly if the texture is bound with a resource heap and the PSO layout contains the `BarrierFlags::StorageTexture` [barrier flags](https://github.com/LukasBanana/LLGL/blob/master/include/LLGL/PipelineLayoutFlags.h#L536).

Having said that, I just fixed it to also work correctly with dynamic resource bindings and also added a new function to the command buffer interface for fine-tuning. I will submit these fixes as soon as I have it implemented for Vulkan, D3D12 is already done.

In the meantime, you'd have to bind this texture with a `ResourceHeap` and add `barrierFlags = LLGL::BarrierFlags::StorageTexture` to the respective PSO layout.

st0rmbtw:
Okay, thank you!

st0rmbtw:
> In the meantime, you'd have to bind this texture with a `ResourceHeap` and add `barrierFlags = LLGL::BarrierFlags::StorageTexture` to the respective PSO layout.

It's solved the issue for the `D3D12` backend, but hasn't solved for the `Vulkan`.

LukasBanana:
[...]

PR #163, opened 2025-02-18
Learning and minor corrections

CyberKenneth:
added a note around 309 with a possible issue with sync and saw a , where a ; should be i believe at 240

### DIFF
--- a/sources/Renderer/Vulkan/Texture/VKTexture.cpp
+++ b/sources/Renderer/Vulkan/Texture/VKTexture.cpp
@@ -306,17 +306,6 @@ VkImageLayout VKTexture::TransitionImageLayout(
     if (flushBarrier)
         context.FlushBarriers();
     return oldLayout;
-    // CRITICAL ISSUE Possible: This function doesn't update any internal state to track the new layout.
-    // Without tracking the layout internally, subsequent calls to GetVkImageLayout() may return incorrect information.
-    // This can lead to synchronization errors, validation layer errors, and potential crashes.
-    // Consider adding:
-    // if (flushBarrier) {
-    //     context.FlushBarriers();
-    //     currentLayout_ = newLayout;  // Only update layout when barriers are flushed
-    // } else {
-    //     pendingLayout_ = newLayout;  // Otherwise track that there's a pending transition
-    //     hasPendingTransition_ = true;
-    // }
 }
 
 VkImageLayout VKTexture::TransitionImageLayout(
--- a/sources/Renderer/Vulkan/Texture/VKTexture.cpp
+++ b/sources/Renderer/Vulkan/Texture/VKTexture.cpp
@@ -237,7 +237,7 @@ void VKTexture::CreateImageView(
     VkImageSubresourceRange subresourceRange;
     {
         subresourceRange.aspectMask     = VKImageUtils::GetExclusiveVkImageAspect(viewVkFormat); //TODO: allow stencil-component to be selected
-        subresourceRange.baseMipLevel   = textureViewDesc.subresource.baseMipLevel,
+        subresourceRange.baseMipLevel   = textureViewDesc.subresource.baseMipLevel;
         subresourceRange.levelCount     = textureViewDesc.subresource.numMipLevels;
         subresourceRange.baseArrayLayer = textureViewDesc.subresource.baseArrayLayer;
         subresourceRange.layerCount     = textureViewDesc.subresource.numArrayLayers;
@@ -306,6 +306,17 @@ VkImageLayout VKTexture::TransitionImageLayout(
     if (flushBarrier)
         context.FlushBarriers();
     return oldLayout;
+    // CRITICAL ISSUE Possible: This function doesn't update any internal state to track the new layout.
+    // Without tracking the layout internally, subsequent calls to GetVkImageLayout() may return incorrect information.
+    // This can lead to synchronization errors, validation layer errors, and potential crashes.
+    // Consider adding:
+    // if (flushBarrier) {
+    //     context.FlushBarriers();
+    //     currentLayout_ = newLayout;  // Only update layout when barriers are flushed
+    // } else {
+    //     pendingLayout_ = newLayout;  // Otherwise track that there's a pending transition
+    //     hasPendingTransition_ = true;
+    // }
 }
 
 VkImageLayout VKTexture::TransitionImageLayout(


sources/Renderer/Vulkan/Texture/VKTexture.cpp
```diff
@@ -306,6 +306,17 @@ VkImageLayout VKTexture::TransitionImageLayout(
     if (flushBarrier)
         context.FlushBarriers();
     return oldLayout;
+    // CRITICAL ISSUE Possible: This function doesn't update any internal state to track the new layout.
```
LukasBanana, line 309: `VKDeviceImage::TransitionImageLayout` tracks the state of its image layout. But I do have to acknowledge that state tracking is exactly the part of the Vulkan backend that needs much improvement.
Instead of only adding a larger comment without improving that part of the code, however, I would prefer to keep those change requests to a new ticket for better issue tracking instead of a code PR.

sources/Renderer/Vulkan/Texture/VKTexture.cpp
```diff
@@ -237,7 +237,7 @@ void VKTexture::CreateImageView(
     VkImageSubresourceRange subresourceRange;
     {
         subresourceRange.aspectMask     = VKImageUtils::GetExclusiveVkImageAspect(viewVkFormat); //TODO: allow stencil-component to be selected
-        subresourceRange.baseMipLevel   = textureViewDesc.subresource.baseMipLevel,
+        subresourceRange.baseMipLevel   = textureViewDesc.subresource.baseMipLevel;
```
LukasBanana, line 240: Good catch. Although semantically equivalent, which actually surprised me, since the comma operator can cause some serious headache, I do appreciate the attention to detail :)

LukasBanana (changes requested):

sources/Renderer/Vulkan/Texture/VKTexture.cpp
CyberKenneth: Thank you! I am a university student I love debugging so I got used to looking for these little things. Had one of the best and strictest professors and TA combos. if you said c99,cpp11, c17 etc.  if you said it you were expected to strictly adhere to the standard and got docked harshly.

sources/Renderer/Vulkan/Texture/VKTexture.cpp
CyberKenneth: Sorry about that. In my previous coursework, we were taught to add inline comments about potential improvements before making production code changes. I'm still learning the ropes of open-source contribution on GitHub—this was my 3rd code change or so, and I wasn't sure about the ticket system here.
I tried reading through the project documentation you shared, but it didn't help me understand the ticket creation process. Because of that, I defaulted to what I know from school, but I'm eager to understand the right approach for this project.
Would you mind guiding me through the best way to create a ticket for this improvement? I didn't mean for the suggested code changes to sound rushed or anything. I genuinely want to help improve the Vulkan backend's state tracking, but I want to do it the right way.
The goal is to address potential synchronization issues and make the layout transition tracking more robust. I'm open to feedback on how to approach this most effectively for the project. PS I was told to look for a contributing.md, but I didn't see it or anything in conventions for this type of thing.( Again, sorry if the issue tone is not hostile or negative I mean to be positive and genuinely want to help. Have ASD and sometimes come off the wrong way in person or text ). Let me know how to proceed :)

CyberKenneth (commented):
I remember commenting on the layout tracking issue because not updating the internal state after transitioning the image layout could lead to synchronization problems. Without correctly tracking the layout (like currentLayout_ or pendingLayout_), subsequent operations might use the wrong layout, causing rendering artifacts and Vulkan validation errors. This could be the reason behind the issue #145 I believe was it. This was my tiny suggestion to see if it helped while it's not the entire file and stuff it is something that in theory May improve performance or decrease artifacts and if that was the case then we could go further

sources/Renderer/Vulkan/Texture/VKTexture.cpp
LukasBanana: No worries, I didn't take any offense and hopefully the same goes the other way. Some details gets lost in pure text communication.

The way your coursework goes about code comments might be perfectly valid. As for this project, I think it makes sense to make use of the issue tracker for these cases. The best way to provide a contribution with pointing out critical issues would be to add a new unit test that proves that something is not working correctly. This way you'd be providing not only a comment about something that is wrong but also an effective way for everyone to reproduce the issue.

CyberKenneth: Thank you for the advice. Do you have any videos, books, or sites about this? I am a junior, so the only real tests I have dealt with were the ones TAs and professors made. This would be the first real-time isolation, something like this. I found this small thing here by accident.  Read documents about Barriers and saw this. So, any advice helps.

CyberKenneth:
removed the comment  in

 ````c++ VkImageLayout VKTexture::TransitionImageLayout(
    VKCommandContext&           context,
    VkImageLayout               newLayout,
    bool                        flushBarrier)
{
    const TextureSubresource fullSubresource{ 0, numArrayLayers_, 0, numMipLevels_ };
    VkImageLayout oldLayout = image_.TransitionImageLayout(context, GetVkFormat(), newLayout, fullSubresource);
    if (flushBarrier)
        context.FlushBarriers();
    return oldLayout;
    // CRITICAL ISSUE Possible: This function doesn't update any internal state to track the new layout.
    // Without tracking the layout internally, subsequent calls to GetVkImageLayout() may return incorrect information.
    // This can lead to synchronization errors, validation layer errors, and potential crashes.
    // Consider adding:
    // if (flushBarrier) {
    //     context.FlushBarriers();
    //     currentLayout_ = newLayout;  // Only update layout when barriers are flushed
    // } else {
    //     pendingLayout_ = newLayout;  // Otherwise track that there's a pending transition
    //     hasPendingTransition_ = true;
    // }
}

CyberKenneth:
> > I closed by accident
> 
> Can you re-open the PR or do you need assistance with that? Can't tell if you have the permissions. Otherwise, feel free to open a new one.

I do not have the perms sorry

CyberKenneth:
oh i do apparently sorry.

LukasBanana (approved):
Thank you for your first contribution and bug fix.

CyberKenneth:
My pleasure!

LukasBanana:
I see you closed the PR. I approved it, but wanted to wait until the CI jobs completed before I merge the changes.

CyberKenneth:
> I see you closed the PR. I approved it, but wanted to wait until the CI jobs completed before I merge the changes.

yes sorry still new to this site I did most work on a home server. so thanks

PR #167, opened 2025-02-20
allow using vulkan on macos

coco875:
It's just allow macos to use vulkan when LLGL_BUILD_RENDERER_VULKAN are enable, I only test it here https://github.com/coco875/test-llgl/tree/216881a74cdfa4202bd607d09aa7253ba4920c43 and it compile but can't try completly because imgui need more setup.

### DIFF
--- a/.github/workflows/ci_macos.yml
+++ b/.github/workflows/ci_macos.yml
@@ -2,7 +2,7 @@ name: Apple macOS
 
 on:
   push:
-    branches: [ "master" ]
+    branches: [ "*" ]
   pull_request:
     branches: [ "master" ]
 
--- a/.github/workflows/ci_macos.yml
+++ b/.github/workflows/ci_macos.yml
@@ -53,7 +53,7 @@ jobs:
 
     - name: Upload Binaries
       uses: actions/upload-artifact@v4
-      if: ${{ always() && matrix.ext == 'Full' && matrix.lib == 'Shared' && matrix.config == 'Release' }}
+      if: ${{ matrix.ext == 'Full' && matrix.lib == 'Shared' && matrix.config == 'Release' }}
       with:
         name: LLGL-macOS-${{ matrix.config }}-x86_64
         path: |
--- a/.github/workflows/ci_macos.yml
+++ b/.github/workflows/ci_macos.yml
@@ -29,7 +29,7 @@ jobs:
     
     - name: Install dependencies
       run: >
-        brew install vulkan-sdk
+        brew install vulkan-tools
 
     - name: Configure CMake
       run: >
--- a/.github/workflows/ci_macos.yml
+++ b/.github/workflows/ci_macos.yml
@@ -53,7 +53,7 @@ jobs:
 
     - name: Upload Binaries
       uses: actions/upload-artifact@v4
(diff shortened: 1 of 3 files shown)


LukasBanana (changes requested):
A bit more details in the description would be highly appreciated.

Were you able to test this? The CI builds won't pick this up because of a few other places this has to be enabled:
1. CMake needs to be updated to allow `LLGL_BUILD_RENDERER_VULKAN` on Mac: [CMakeLists.txt:458](https://github.com/LukasBanana/LLGL/blob/master/CMakeLists.txt#L458)
2. Mac CI script should enable this (somwhere here [ci_macos.yml:34](https://github.com/LukasBanana/LLGL/blob/master/.github/workflows/ci_macos.yml#L34)), but this will require a GitHub Action dependency to MoltenVK.

coco875:
I test it in https://github.com/coco875/test-llgl/tree/metal-test (where I manually apply the commit). I will try tommorrow to add it to the ci

coco875:
I try to enable ci and add LLGL_BUILD_RENDERER_VULKAN=ON and it work, don't understand what do you mean by "CMake needs to be updated to allow LLGL_BUILD_RENDERER_VULKAN on Mac"

.github/workflows/ci_macos.yml
```diff
@@ -2,14 +2,15 @@ name: Apple macOS
 
 on:
   push:
-    branches: [ "master" ]
+    branches: [ "*" ]
```
LukasBanana, line 5: The GitHub actions take up a good amount of time and I think failed jobs will always show up in the badge on the front  page, which is not really necessary for experimental/side branches. So I'd like to keep it to the `master` branch for now. If we open this up to other branches, I would rather do it in one commit for all CI jobs.

coco875: Ok, it was just to experiment on my fork, I will revert it

.github/workflows/ci_macos.yml
```diff
@@ -46,7 +53,7 @@ jobs:
 
     - name: Upload Binaries
       uses: actions/upload-artifact@v4
-      if: matrix.lib == 'Static' && matrix.config == 'Release'
+      if: ${{ always() && matrix.ext == 'Full' && matrix.lib == 'Shared' && matrix.config == 'Release' }}
```
LukasBanana, line 56: I forgot about the `always()` condition for a moment, but this was to make sure the test report is always uploaded even when (or especially when rather) the test run failed. As long as there is no step for the Testbed, we don't need this here, though.

Please also revert the `Shared`-build change. The idea to have the static build to be uploaded is to reduce the number of binary files that have to be signed in order to run on another Mac. Code signing is currently not handled at all by any script in this repository (not sure how that would work on a build script anyway), but at least whoever downloads the build artifacts has to sign fewer files to run the examples.

My intention was to only add the missing `matrix.ext == 'Full` check, because otherwise, there's a conflict in the CI step when two build configuration try to upload the same artifact file.

LukasBanana (changes requested):

.github/workflows/ci_macos.yml
coco875: ok ok

LukasBanana (approved):
Looks. Thanks for your contribution. I'll merge as soon as the CI builds complete.

PR #175, opened 2025-03-11
fix create a vulkan instance with moltenvk

coco875:
fix `VK_ERROR_INCOMPATIBLE_DRIVER` when `vkCreateInstance` is call with moltenvk. (based on this https://stackoverflow.com/questions/58732459/vk-error-incompatible-driver-with-mac-os-and-vulkan-moltenvk)

### DIFF
--- a/sources/Renderer/Vulkan/VKRenderSystem.cpp
+++ b/sources/Renderer/Vulkan/VKRenderSystem.cpp
@@ -821,15 +821,20 @@ void VKRenderSystem::CreateInstance(const RendererConfigurationVulkan* config)
         );
     };
 
+    /* Setup Vulkan instance descriptor */
+    VkInstanceCreateInfo instanceInfo = {};
+
     for (const VkExtensionProperties& prop : extensionProperties)
     {
         const VKExtSupport extSupport = GetVulkanInstanceExtensionSupport(prop.extensionName);
         if (IsVKExtSupportIncluded(extSupport))
             extensionNames.push_back(prop.extensionName);
+#ifdef VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME
+        if (strcmp(prop.extensionName, VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME) == 0)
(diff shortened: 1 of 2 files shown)


sources/Renderer/Vulkan/VKRenderSystem.cpp
```diff
@@ -821,16 +821,21 @@ void VKRenderSystem::CreateInstance(const RendererConfigurationVulkan* config)
...
+
     for (const VkExtensionProperties& prop : extensionProperties)
     {
         const VKExtSupport extSupport = GetVulkanInstanceExtensionSupport(prop.extensionName);
         if (IsVKExtSupportIncluded(extSupport))
             extensionNames.push_back(prop.extensionName);
+#ifdef VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME
+        if (strcmp(prop.extensionName, VK_KHR_PORTABILITY_ENUMERATION_EXTENSION_NAME) == 0)
+            instanceInfo.flags |= VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR;
+            extensionNames.push_back(prop.extensionName);
```
LukasBanana, line 835: If adding the extension name is meant to be part of the `if`-statement, then that statement must have a compound statement, i.e. it's missing the curly braces `{`/`}`.
Besides that, can you please replace the statement that adds the extension name by adding it to [`IsVulkanInstanceExtOptional()`](https://github.com/LukasBanana/LLGL/blob/master/sources/Renderer/Vulkan/Ext/VKExtensionRegistry.cpp#L78C13-L78C40) instead. This will add the name in the conditional block above. We can keep the rest as a special case to add this bit field.

LukasBanana (changes requested):

sources/Renderer/Vulkan/VKRenderSystem.cpp
coco875: (oh forget to add {}, I prefer always put {} but like it look like you prefer remove them for line I iniatially don't add them but forgot to put it when I add the second line)

LukasBanana:
Thanks for the fix. Looks good to me.

PR #182, opened 2025-03-29
Create a true copy of the initial blob in D3D12PipelineCache

st0rmbtw:
The `Blob`'s `CreateCopy` method doesn't actually create a copy, it only copies the pointer and the size, that can cause a dangling pointer issue when the initial blob is destroyed.

Example:
```cpp
LLGL::PipelineCache* ReadPipelineCache(const std::string& path, bool& hasInitialCache) {
    LLGL::Blob pipelineCacheBlob = LLGL::Blob::CreateFromFile(path);
    if (pipelineCacheBlob) {
        hasInitialCache = true;
    }

    return m_context->CreatePipelineCache(pipelineCacheBlob);
    // pipelineCacheBlob gets destroyed here
}

int main() {
    ...
    bool hasInitialCache = false;
    LLGL::PipelineCache* pipelineCache = ReadPipelineCache("GraphicsPSO.D3D12.cache", hasInitialCache);
    context->CreatePipelineState(pipelineDesc, pipelineCache); // This line will cause a segfault
    ...
}
```

### DIFF
--- a/sources/Renderer/Direct3D12/RenderState/D3D12PipelineCache.cpp
+++ b/sources/Renderer/Direct3D12/RenderState/D3D12PipelineCache.cpp
@@ -13,7 +13,7 @@ namespace LLGL
 
 
 D3D12PipelineCache::D3D12PipelineCache(const Blob& initialBlob) :
-    initialBlob_ { initialBlob ? Blob::CreateCopy(initialBlob.GetData(), initialBlob.GetSize()) : Blob{} }
+    initialBlob_ { initialBlob ? Blob::CreateWeakRef(initialBlob.GetData(), initialBlob.GetSize()) : Blob{} }
 {
 }
 


LukasBanana:
I think the actual bug is that the branches the condition at [Blob.cpp:117](https://github.com/LukasBanana/LLGL/blob/4adb25c989e37ac23225ac28a9752a1adadfef52/sources/Core/Blob.cpp#L117) is leads to are mixed up. When `isWeakRef` is true, an internal instance of `InternalUnmanagedBlob` should be created and not `InternalVectorBlob`.

LukasBanana:
I went ahead and fixed this mix-up. Thanks for bringing this up.

For context: we still want to create a copy of the blob, it was only that the wrong internal representation was instantiated under the `isWeakRef` condition. Weak referenced blobs are only intended if you want to avoid a copy operation when you can guarantee the lifetime of the blob will outlive the object/PSO-cache that's referencing it.

PR #184, opened 2025-04-02
Allow to not use STL (Part 1)

soufianekhiat:
Allow the replacement of STL use with for instance EASTL or anything else with the same interface.

Split all includes on differents files.

cf. LLGL/Container/...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -143,7 +143,7 @@ option(LLGL_ENABLE_CHECKED_CAST "Enable dynamic checked cast (only in Debug mode
 option(LLGL_ENABLE_DEBUG_LAYER "Enable renderer debug layer (for both Debug and Release mode)" ON)
 option(LLGL_ENABLE_EXCEPTIONS "Enable C++ exceptions" OFF)
 
-option(LLGL_PREFER_STL_CONTAINERS "Prefers C++ STL containers over custom containers, e.g. std::vector over SmallVector<T>" OFF)
+option(LLGL_PREFER_STL_CONTAINERS "Prefers C++ STL containers over custom containers, e.g. vector over SmallVector<T>" OFF)
 
 option(LLGL_BUILD_STATIC_LIB "Build LLGL as static library" OFF)
 option(LLGL_BUILD_TESTS "Include test projects" OFF)
--- a/examples/Cpp/Animation/Example.cpp
+++ b/examples/Cpp/Animation/Example.cpp
@@ -54,7 +54,7 @@ class Example_Animation : public ExampleBase
         float                   frameInterpolator   = 0.0f;
     };
 
-    std::vector<Ball>           balls;
+    vector<Ball>                balls;
 
     const Gs::Vector2f          gridPosFrames[15]
     {
@@ -106,7 +106,7 @@ private:
         vertexFormat.SetStride(sizeof(TexturedVertex));
 
         // Load 3D models
-        std::vector<TexturedVertex> vertices;
+        vector<TexturedVertex> vertices;
         meshStairsTop       = LoadObjModel(vertices, "PenroseStairs-Top.obj");
         meshStairsBottom    = LoadObjModel(vertices, "PenroseStairs-Bottom.obj");
         meshBall            = LoadObjModel(vertices, "IcoSphere.obj");
--- a/examples/Cpp/ClothPhysics/Example.cpp
+++ b/examples/Cpp/ClothPhysics/Example.cpp
@@ -144,9 +144,9 @@ public:
 
     // Generates the grid geometry for the cloth with triangle strip topology
     void GenerateClothGeometry(
-        std::vector<ParticleBase>&  verticesBase,
-        std::vector<Gs::Vector4f>&  verticesPos,
-        std::vector<std::uint32_t>& indices)
+        vector<ParticleBase>&  verticesBase,
+        vector<Gs::Vector4f>&  verticesPos,
+        vector<std::uint32_t>& indices)
     {
         const auto invSegsU = 1.0f / static_cast<float>(clothSegmentsU);
         const auto invSegsV = 1.0f / static_cast<float>(clothSegmentsV);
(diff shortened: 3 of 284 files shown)


LukasBanana:
Thank you for showing interest in LLGL. I've heard of the EASTL a long time ago when I was still a student assistant but never had a chance to take a closer look at it. I somehow assumed it was a 100% compatible implementation of the C++ STL as if it lived in the same `std::` namespace.

Considering that this is quite an intrusive change, what is your use case for LLGL that requires EASTL?
Is this first and foremost to make the public interface avoid STL containers? For that matter I wanted to replace the STL containers at least in the public interface with `SmallVector`, `StringView`, and `UTF8String` eventually since I know the STL containers don't mix well across DLL boundaries, but I haven't found the time yet.
And did you have a chance to run a performance comparison with LLGL that compares the STL and EASTL implementations?

I embraced the use of explicit namespaces, especially in public interfaces, so including `<LLGL/LLGL.h>` won't interfere with your own project's use of `std::` types.
I'm afraid aliasing the `std` namespace to point the compiler to `eastl` is not a real option but perhaps you have an idea if there's a less intrusive solution?

If EASTL is fully compatible or compatible enough, would it be feasible to just have wrapper headers under `external/EASTL/` that do something like this:
```cpp
// "vector" file
#include <eastl/vector.h>
namespace std
{
template <typename T, typename Allocator = eastl::EASTLAllocatorType>
using vector = eastl::vector<T, Allocator>;
}
```
This would effectively substitute the existing uses of STL types with EASTL but no or only a few code changes throughout LLGL would be necessary. This of course requires that the client project directs all standard headers to this folder with all the wrapper headers. I haven't done something like this before, but would like to explore some other options first.

soufianekhiat:
For performance I trusted the benchmark of EA:
https://github.com/electronicarts/EASTL/blob/master/doc/Benchmarks.md
But didn't profile LLGL with EASTL.

I don't use EASTL only for performance but mostly about his single interface sharable as dll or static lib without compiler difference of implementation etc.
And I always override the allocator for profiling and memory tracking.

Which explain my use of:
```cpp
#ifdef LLGL_REPLACE_STD_VECTOR
template < typename Type >
using vector = LLGL_CUSTOM_VECTOR;
#else
template < typename Type >
using vector = std::vector< Type >;
#endif
```

I read you don't want to increase the number of dependencies so I choose that implementation to move the integration of EASTL on the hand of the dev not to LLGL.

Ideally I wanted EASTL for vector, set, stack and robin_map for map.

If you accept that Part 1 I will change the other one std::share_ptr, std::unique_ptr, ...

To reduce the intruisive changes we could maybe use the strategy of Dear ImGui like having a config.h that allow us to override internal types.

LukasBanana:
I tried to build your branch with the EASTL and here are a few things I came across:
1. At least `eastl::basic_string` is incompatible with `std::basic_string` because it doesn't provide a type traits template parameter. This would need another constructor in `LLGL::BasicStringLiteral` because this one does not implicitly convert from `eastl::string` to its own type:
```cpp
//! Initializes the string literal with another templated string class. This would usually be std::basic_string.
template <template <class, class, class> class TString>
BasicStringLiteral(const TString<TChar, Traits, Allocator>& str)
```
2. There doesn't seem to be an implementation of _regex_ nor does the `eastl::basic_string` class implicitly convert to `std::basic_string` to make it compatible. There's only two files in the examples and unit tests that use regex, so this wouldn't be a big issue, but the incompatibility concerns me under the consideration that EASTL attempts to align with the STL API.
3. This incompatibility continues with other standard library classes such as `ifstream`, which is also not used a lot, in fact I removed as many of the I/O streams, but it's semi-optimal IMO.
4. EASTL seems to be dependent on [EABase](https://github.com/electronicarts/EABase) even when I add `#define EASTL_EABASE_DISABLED` before including the headers. While you made the change generic to allow any library to replace the STD containers, the feasibility of using EASTL specifically adds two new dependencies at once and I'm questioning if this is worth the overhaul of all STD uses in this project.

I stopped trying to build after resolving some of the compiler errors (I have about 50 remaining), which mostly seem to stem from interoperability issues with `eastl::basic_string` and other STD library functions such as `fstream`, `regex`, replacing `std::to_string()` with `east::to_string` was an easy fix.

> But didn't profile LLGL with EASTL.

Did you at least build LLGL with EASTL? Perhaps I missed something as I'm not experienced with the EASTL.

> but mostly about his single interface sharable as dll or static lib without compiler difference of implementation etc.

That is indeed a good reason to use such libraries. As I mentioned, this is on my TODO list (I should probably put it onto the public TODO list here on GitHub) to replace all STL containers in the public interface with LLGL's own containers for that matter.

As for replacing all uses of the STD containers in the entire project, I am not quite convinced this has priority for LLGL or is pushing LLGL forward significantly. There are too many other construction sites in this project that need improvement such as Vulkan's resource state tracking, its memory allocator, multi-threading support and more.

If you are starting with a project that shows the need for custom memory allocators for LLGL's containers we can go from there. Until then, I'd like to focus on the public interface only, preferably using LLGL's own containers and also seeing some project or example as a proof of concept why such large changes are necessary.

soufianekhiat:
Indeed the STL and EASTL had some differences for instance the allocators with by design a support for aligned allocators.
https://github.com/electronicarts/EASTL/blob/master/doc/Design.md#allocator-design
EASTL do not try to mimic the stl 1 to 1, but target better debug build perf, performance, memory alignment, ....
For the allocator being compatible we would need an interface which is an union of the stl and eastl.

I built with vector/map/set changes to EASTL version.
I try to have it transparent for LLGL and move the complexity to the user and not the API.

To have it working you need some change like enable some traits on EASTL with:
`#define EASTL_STD_ITERATOR_CATEGORY_ENABLED 1`
before including EASTL/vector.h.

After testing OpenGL Backend (never use it) I did some change to have it building.
I planned to checked for string/stream later, I make sure the CI build and the CL do not reference explictly eastl at all, just allow a replacement, I could hide eastl behind my own interface if needed.

Currently my minimal dependencies stack for working is:
* [EASTL](https://github.com/electronicarts/EASTL/)
* [mi_malloc](https://github.com/microsoft/mimalloc)

For a proper memory managements in CPU. Currently on LLGL (due to the STL usage and their default allocator) I cannot track the memory usage and when in the frame memory spikes could happen.

I'm totally fine with "LLGL's own containers" but STL as is is not viable, I need to know who, how and when my memory is used. To know is I didn't used an API properly or an API is buggy etc.
I like the approach of stb or Dear ImGui with a config.h file or a preprocessor to override all internal allocation, which allow me to tag and track their allocation.

I totally understand the read of this PR as intruisiveness, I don't see it as intruisive but more typedef-ness who allow me to keep my changes local more easily if that PR didn't land.

Feel free to close that PR.

soufianekhiat:
Alternative proposition:
https://github.com/LukasBanana/LLGL/pull/186

LukasBanana:
Closing this in favor of the new PR.

PR #186, opened 2025-04-06
Allow custom allocator for STL container

soufianekhiat:
cf. Title.

Look like lot of changes, but in practice it's just a typedef and changing includes.

I didn't update the uses of STL on Tests.

### DIFF
--- a/include/LLGL/Container/STLAllocator.h
+++ b/include/LLGL/Container/STLAllocator.h
@@ -1,0 +1,34 @@
+/*
+ * STL Allocator.h
+ *
+ * Copyright (c) 2015 Lukas Hermanns. All rights reserved.
+ * Licensed under the terms of the BSD 3-Clause license (see LICENSE.txt).
+ */
+
+#ifndef LLGL_STL_ALLOCATOR_H
+#define LLGL_STL_ALLOCATOR_H
+
+#ifndef LLGL_CUSTOM_ALLOCATOR
+#include <memory>
+#endif
+
+namespace LLGL{
+
+#ifndef LLGL_CUSTOM_ALLOCATOR
+
+template < typename Type >
+using llgl_allocator = std::allocator< Type >;
+
+#else
+
+// Is up to the user to implement a custom allocator 'llgl_allocator'
+
+#endif
+
+} // /namespace LLGL
+
+#endif
+
+
+
+// ================================================================================
--- a/include/LLGL/Container/Set.h
+++ b/include/LLGL/Container/Set.h
@@ -1,0 +1,25 @@
+/*
+ * Set.h
+ *
+ * Copyright (c) 2015 Lukas Hermanns. All rights reserved.
+ * Licensed under the terms of the BSD 3-Clause license (see LICENSE.txt).
+ */
+
+#ifndef LLGL_SET_H
+#define LLGL_SET_H
+
+#include <LLGL/Container/STLAllocator.h>
+
+#include <set>
+#include <utility>
+
+namespace LLGL{
+
+template < typename Type, typename Less = std::less< Type > >
+using set = std::set< Type, Less, llgl_allocator< Type > >;
+
+} // /namespace LLGL
+
+#endif
+
+// ================================================================================
--- a/include/LLGL/Container/StringView.h
+++ b/include/LLGL/Container/StringView.h
@@ -10,7 +10,7 @@
 
 
 #include <LLGL/Export.h>
-#include <string>
+#include <LLGL/Container/String.h>
 #include <cstddef>
 #include <iterator>
 #include <algorithm>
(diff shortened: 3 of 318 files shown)


LukasBanana:
Hi again. I thought a little bit about this and I can see the necessity for custom allocators. I appreciate that you kept the formatting intact but here is a proposal to keep things nicely separated, compatible, and keep all indentation - also because I'm a bit nitpicky about naming :)

If we are about to abstract the standard containers to allow custom allocators and also make them replaceable with third party template libraries such as EASTL, why don't we keep the same names but with different pattern, i.e. keep the convention as with other LLGL containers such as `SmallVector`. To distinguish them from the custom containers, since they are just acting as "proxies" for standard containers, let's put them into a new public namespace named `STL` and their names match their `std::` models but in _PascalCase_ as with LLGL's types. For example, `std::vector` becomes `LLGL::STL::Vector` but the outer scope `LLGL::` is implicit inside LLGL. This way there's no indentation changes anywhere with your PR. It would also ensure that we don't have `using map = std::map` and friends since I don't think it's a good idea to move the std containers into LLGL's scope; if someone has `using namespace LLGL` it might (even if in rare cases) conflict with their use of the C++ STL if their code depends on `map` not being in global scope.

I like the idea of having each container in a separate header to ensure we don't have unnecessary include dependencies when only a single container is needed, since they tend to be very long already. If we go this route, I would also suggest to move all those container wrappers into a new folder under `LLGL/STL/` instead of `LLGL/Container/`. Your choice makes sense, too, but I like to keep this separate from LLGL's own container classes.

Looking forward to hear what you think.

include/LLGL/Container/StringView.h
```diff
@@ -114,7 +114,7 @@ class LLGL_EXPORT BasicStringView
 
         /**
         \brief Returns the raw pointer of this string view.
-        \remarks This does not necessarily point to a NUL-terminated string like \c std::string does.
+        \remarks This does not necessarily point to a NUL-terminated string like \c string does.
```
LukasBanana, line 117: Regarding the changes in the comments, I don't think these will be necessary. I presume you used a _find & replace_ function, but since these comments make an example under the assumption of using the C++ STL by default, I think we should keep them.

include/LLGL/Container/STLAllocator.h
```diff
@@ -0,0 +1,34 @@
...
+#ifndef LLGL_CUSTOM_ALLOCATOR
+#include <memory>
+#endif
+
+namespace LLGL{
+
+#ifndef LLGL_CUSTOM_ALLOCATOR
+
+template < typename Type >
+using llgl_allocator = std::allocator< Type >;
```
LukasBanana, line 20: If we go the route with the `STL` namespace, I would also just put it there and call it with the same naming convention as the other aliases, i.e. `STL::Allocator`.

include/LLGL/Container/STLAllocator.h
```diff
@@ -0,0 +1,34 @@
...
+
+#else
+
+// Is up to the user to implement a custom allocator 'llgl_allocator'
```
LukasBanana, line 24: So if a user defines their own allocator, this would have to be provided not only for their project before their `#include <LLGL/LLGL.h>` directive but also when they build LLGL, since the rest of the library must make use of the same allocator, right?

You mentioned in your previous PR that you liked the concept of ImGui's `config.h` file; would this come into play here as well?

I presume it would also be desirable to opt-in a custom allocator that acts as a wrapper to make it swappable at runtime, so that you can launch your app with different memory allocation policies at launch without rebuilding LLGL.

include/LLGL/Container/Set.h
```diff
@@ -0,0 +1,25 @@
...
+
+#ifndef LLGL_SET_H
+#define LLGL_SET_H
+
+#include <LLGL/Container/STLAllocator.h>
+
+#include <set>
+#include <utility>
+
+namespace LLGL{
```
LukasBanana, line 16: We can look into getting a `.clang-format` up an running as you suggested. I am personally not a huge fan of auto-formatters, but I can see why it makes sense for public projects like this one. Until then, would you mind changing the code indentation to [Allman](https://en.wikipedia.org/wiki/Indentation_style#Allman) style?

LukasBanana (changes requested):

include/LLGL/Container/STLAllocator.h
soufianekhiat: Runtime swappable is an extra feature, having it customization could be a good first step.

soufianekhiat:
Let me know if the changes work for you.

LukasBanana (changes requested):
Before I continue through this huge PR, I think we should first trim it down by removing all the changes in the comment sections as I mentioned at a few places. To summarize my change requests:
1. Please use PascalCase for the new typenames, i.e. `STL::Vector` for `std::vector`, `STL::String` for `std::string` etc. This makes it clear we're dealing with new typenames that are part of LLGL and may or may not refer to the C++ standard library depending on settings (or at least once EASTL or equivalents are hooked up).
2. Please remove all mentions of `STL::` from the comment sections. The mentions of "string" and "vector" don't need to mention specific types but are intended as general reference to their purpose. I think this has the potential to trim down this large PR.

I'm okay with `using LLGL::STL::Map;` etc. in the example projects. I assume you don't have any other `using` or `using namespace` in any other files of the library.

PR #187, opened 2025-04-30
Reorder Metal version checks

st0rmbtw:
This function always returns 101 because the check for that version goes first. Instead it should check higher versions first.

### DIFF
--- a/sources/Renderer/Metal/MTFeatureSet.mm
+++ b/sources/Renderer/Metal/MTFeatureSet.mm
@@ -21,23 +21,26 @@ namespace LLGL
 // Converts the specified Metal feature set into a fixed version number
 static int FeatureSetToVersion(MTLFeatureSet fset)
 {
-    #if MAC_OS_X_VERSION_MIN_REQUIRED >= MAC_OS_X_VERSION_10_11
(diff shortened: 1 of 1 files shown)


sources/Renderer/Metal/MTFeatureSet.mm
```diff
@@ -21,24 +21,27 @@
...
-        return 102; // 1.2
-    #endif
-    #if MAC_OS_X_VERSION_MIN_REQUIRED >= MAC_OS_X_VERSION_10_13
-    if (fset >= MTLFeatureSet_macOS_GPUFamily1_v3)
-        return 103; // 1.3
-    #endif
     #if MAC_OS_X_VERSION_MIN_REQUIRED >= MAC_OS_X_VERSION_10_14
     if (fset >= MTLFeatureSet_macOS_GPUFamily1_v4)
         return 104; // 1.4
     if (fset >= MTLFeatureSet_macOS_GPUFamily2_v1)
```
LukasBanana, line 27: I think this one should also be moved up.

st0rmbtw: Ah, yes, you're right, didn't notice that one

LukasBanana (approved):

PR #188, opened 2025-04-30
Use Metal's D24UNormS8UInt pixel format

st0rmbtw:
When debugging with Xcode it complains about swap chain pixel format mismatch. That's because of the `ToMTLPixelFormat` function which returns wrong Metal pixel format for the `D24UNormS8UInt` format.

This PR associates LLGL's `D24UNormS8UInt` pixel format with Metal's `MTLPixelFormatDepth32Float_Stencil8` if target OS is not iOS


LukasBanana (approved):
Looks good to me.

PR #190, opened 2025-05-06
Fix misspelled GL extension macro

st0rmbtw:


LukasBanana:
Good catch. Thank you.

PR #193, opened 2025-05-10
Add Wayland support

st0rmbtw:
Added Wayland support on Linux.

This PR makes the `Vulkan` and `OpenGL` backends support Wayland surfaces.

P. S.:

I installed Linux for the first time a couple of days ago and I know nothing about Wayland and EGL, so the implementation might be (and probably is) bad.

Please, give me some feedback on how to improve the code quality and the implementation in general.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -176,7 +176,7 @@ if(WIN32)
     option(LLGL_BUILD_RENDERER_DIRECT3D12 "Include Direct3D12 renderer project" ON)
 endif()
 
-if(LLGL_TARGET_PLATFORM STREQUAL "Linux")
+if(UNIX AND NOT APPLE)
     option(LLGL_LINUX_ENABLE_WAYLAND "Enable support for Wayland protocol" OFF)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -176,6 +176,10 @@ if(WIN32)
     option(LLGL_BUILD_RENDERER_DIRECT3D12 "Include Direct3D12 renderer project" ON)
 endif()
 
+if(LLGL_TARGET_PLATFORM STREQUAL "Linux")
+    option(LLGL_LINUX_ENABLE_WAYLAND "Enable support for Wayland protocol" OFF)
+endif()
+
 if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.16")
     option(LLGL_UNITY_BUILD "Build LLGL modules as unity builds in batches. This can speedup compilation time." OFF)
 endif()
@@ -542,6 +546,10 @@ endif()
 
 if(LLGL_UNITY_BUILD)
     ADD_DEFINE(LLGL_UNITY_BUILD)
+endif()
+
+if (LLGL_LINUX_ENABLE_WAYLAND)
+    ADD_DEFINE(LLGL_LINUX_ENABLE_WAYLAND)
 endif()
 
 if(LLGL_WASM_PLATFORM)
--- a/include/LLGL/Backend/OpenGL/Linux/LinuxNativeHandle.h
+++ b/include/LLGL/Backend/OpenGL/Linux/LinuxNativeHandle.h
@@ -9,8 +9,12 @@
 #define LLGL_OPENGL_LINUX_NATIVE_HANDLE_H
 
 
+#include <LLGL/Deprecated.h>
 #include <GL/glx.h>
-#include <EGL/egl.h>
+
+#ifdef LLGL_LINUX_ENABLE_WAYLAND
+    #include <EGL/egl.h>
+#endif
 
 
 namespace LLGL
@@ -20,8 +24,10 @@ namespace OpenGL
 {
 
 enum class RenderSystemNativeHandleType : char {
-    GLX,
+    GLX = 0,
+#ifdef LLGL_LINUX_ENABLE_WAYLAND
     EGL
+#endif
 };
 
 /**
@@ -33,6 +39,8 @@ struct RenderSystemNativeHandle
 {
     union
     {
+        LLGL_DEPRECATED("Deprecated since 0.04b; Use glx instead.")
+        GLXContext context;
         GLXContext glx;
         EGLContext egl;
     };
--- a/include/LLGL/Backend/OpenGL/Linux/LinuxNativeHandle.h
+++ b/include/LLGL/Backend/OpenGL/Linux/LinuxNativeHandle.h
@@ -10,6 +10,7 @@
 
 
 #include <GL/glx.h>
+#include <EGL/egl.h>
 
 
 namespace LLGL
@@ -18,6 +19,10 @@ namespace LLGL
 namespace OpenGL
 {
 
+enum class RenderSystemNativeHandleType : char {
+    GLX,
+    EGL
+};
 
 /**
 \brief GNU/Linux native handle structure for the OpenGL render system.
@@ -26,7 +31,13 @@ namespace OpenGL
 */
 struct RenderSystemNativeHandle
 {
-    GLXContext context;
+    union
+    {
+        GLXContext glx;
+        EGLContext egl;
+    };
+
+    RenderSystemNativeHandleType type;
 };
 
 
--- a/include/LLGL/Backend/OpenGL/Linux/LinuxNativeHandle.h
+++ b/include/LLGL/Backend/OpenGL/Linux/LinuxNativeHandle.h
@@ -40,7 +40,11 @@ struct RenderSystemNativeHandle
         LLGL_DEPRECATED("Deprecated since 0.04b; Use glx instead.")
         GLXContext context;
         GLXContext glx;
+        #if LLGL_EXPOSE_WAYLAND || LLGL_LINUX_ENABLE_WAYLAND
         EGLContext egl;
+        #else
+        void*      egl;
+        #endif
     };
 
     RenderSystemNativeHandleType type;
--- a/include/LLGL/Backend/OpenGL/Linux/LinuxNativeHandle.h
+++ b/include/LLGL/Backend/OpenGL/Linux/LinuxNativeHandle.h
@@ -12,7 +12,7 @@
 #include <LLGL/Deprecated.h>
 #include <GL/glx.h>
 
-#ifdef LLGL_LINUX_ENABLE_WAYLAND
+#if LLGL_EXPOSE_WAYLAND || LLGL_LINUX_ENABLE_WAYLAND
     #include <EGL/egl.h>
 #endif
 
@@ -25,9 +25,7 @@ namespace OpenGL
 
 enum class RenderSystemNativeHandleType : char {
     GLX = 0,
-#ifdef LLGL_LINUX_ENABLE_WAYLAND
     EGL
-#endif
 };
 
 /**
@@ -42,9 +40,7 @@ struct RenderSystemNativeHandle
         LLGL_DEPRECATED("Deprecated since 0.04b; Use glx instead.")
         GLXContext context;
         GLXContext glx;
-    #ifdef LLGL_LINUX_ENABLE_WAYLAND
         EGLContext egl;
-    #endif
     };
 
     RenderSystemNativeHandleType type;
--- a/include/LLGL/Backend/OpenGL/Linux/LinuxNativeHandle.h
+++ b/include/LLGL/Backend/OpenGL/Linux/LinuxNativeHandle.h
@@ -42,7 +42,9 @@ struct RenderSystemNativeHandle
         LLGL_DEPRECATED("Deprecated since 0.04b; Use glx instead.")
         GLXContext context;
         GLXContext glx;
+    #ifdef LLGL_LINUX_ENABLE_WAYLAND
         EGLContext egl;
+    #endif
     };
 
     RenderSystemNativeHandleType type;
--- a/include/LLGL/Backend/OpenGL/Linux/LinuxNativeHandle.h
+++ b/include/LLGL/Backend/OpenGL/Linux/LinuxNativeHandle.h
@@ -23,7 +23,7 @@ namespace LLGL
 namespace OpenGL
 {
 
-enum class RenderSystemNativeHandleType : char {
+enum class RenderSystemNativeType : char {
     GLX = 0,
     EGL
 };
@@ -35,6 +35,8 @@ enum class RenderSystemNativeHandleType : char {
 */
 struct RenderSystemNativeHandle
 {
+    RenderSystemNativeType type;
+
     union
     {
         LLGL_DEPRECATED("Deprecated since 0.04b; Use glx instead.")
@@ -46,8 +48,6 @@ struct RenderSystemNativeHandle
         void*      egl;
         #endif
     };
-
-    RenderSystemNativeHandleType type;
 };
 
 
--- a/include/LLGL/Platform/Linux/LinuxNativeHandle.h
+++ b/include/LLGL/Platform/Linux/LinuxNativeHandle.h
@@ -14,7 +14,7 @@
 
 #include <LLGL/Deprecated.h>
 
-#ifdef LLGL_LINUX_ENABLE_WAYLAND
+#if LLGL_LINUX_ENABLE_WAYLAND
 #include <wayland-client.h>
 #endif
 
@@ -39,7 +39,7 @@ struct X11NativeHandle
     int             screen;
 };
 
-#ifdef LLGL_LINUX_ENABLE_WAYLAND
+#if LLGL_LINUX_ENABLE_WAYLAND
 
 struct WaylandNativeHandle
 {
@@ -49,7 +49,11 @@ struct WaylandNativeHandle
 
 #else
 
-struct WaylandNativeHandle{};
+struct WaylandNativeHandle
+{
+    void* window;
+    void* display;
+};
 
 #endif
 
--- a/include/LLGL/Platform/Linux/LinuxNativeHandle.h
+++ b/include/LLGL/Platform/Linux/LinuxNativeHandle.h
@@ -12,6 +12,8 @@
 #include <X11/Xlib.h>
 #include <X11/Xutil.h>
 
+#include <LLGL/Deprecated.h>
+
 #ifdef LLGL_LINUX_ENABLE_WAYLAND
 #include <wayland-client.h>
 #endif
@@ -61,6 +63,26 @@ enum class NativeHandleType : char
 struct NativeHandle
 {
     union {
+        //! X11 display connection.
+        LLGL_DEPRECATED("Deprecated since 0.04b; Use x11.display instead.")
+        ::Display*      display;
+
+        //! X11 window object.
+        LLGL_DEPRECATED("Deprecated since 0.04b; Use x11.window instead.")
+        ::Window        window;
+
+        //! X11 visual information.
+        LLGL_DEPRECATED("Deprecated since 0.04b; Use x11.visual instead.")
+        ::XVisualInfo*  visual;
+
+        //! X11 colormap object. Used internally by the OpenGL backend.
+        LLGL_DEPRECATED("Deprecated since 0.04b; Use x11.colorMap instead.")
+        ::Colormap      colorMap;
+
+        //! X11 screen index.
+        LLGL_DEPRECATED("Deprecated since 0.04b; Use x11.screen instead.")
+        int             screen;
+
         X11NativeHandle x11;
         WaylandNativeHandle wayland;
     };
--- a/include/LLGL/Platform/Linux/LinuxNativeHandle.h
+++ b/include/LLGL/Platform/Linux/LinuxNativeHandle.h
@@ -12,13 +12,12 @@
 #include <X11/Xlib.h>
 #include <X11/Xutil.h>
 
+#include <wayland-client.h>
 
 namespace LLGL
 {
 
-
-//! Linux native handle structure.
-struct NativeHandle
+struct X11NativeHandle
 {
     //! X11 display connection.
     ::Display*      display;
@@ -34,6 +33,29 @@ struct NativeHandle
 
     //! X11 screen index.
     int             screen;
+};
+
+struct WaylandNativeHandle
+{
+    struct wl_surface* window;
+    struct wl_display* display;
+};
+
+enum class NativeHandleType : char
+{
+    X11 = 0,
+    Wayland = 1
+};
+
+//! Linux native handle structure.
+struct NativeHandle
+{
+    union {
+        X11NativeHandle x11;
+        WaylandNativeHandle wayland;
+    };
+
+    NativeHandleType type;
 };
 
 
--- a/include/LLGL/Platform/Linux/LinuxNativeHandle.h
+++ b/include/LLGL/Platform/Linux/LinuxNativeHandle.h
@@ -93,7 +93,7 @@ struct NativeHandle
         WaylandNativeHandle wayland;
     };
 
-    NativeHandleType type;
+    NativeHandleType type = NativeHandleType::X11;
 };
 
 
--- a/include/LLGL/Platform/Linux/LinuxNativeHandle.h
+++ b/include/LLGL/Platform/Linux/LinuxNativeHandle.h
@@ -12,7 +12,9 @@
 #include <X11/Xlib.h>
 #include <X11/Xutil.h>
 
+#ifdef LLGL_LINUX_ENABLE_WAYLAND
 #include <wayland-client.h>
+#endif
 
 namespace LLGL
 {
@@ -35,11 +37,19 @@ struct X11NativeHandle
     int             screen;
 };
 
+#ifdef LLGL_LINUX_ENABLE_WAYLAND
+
 struct WaylandNativeHandle
 {
     struct wl_surface* window;
     struct wl_display* display;
 };
+
+#else
+
+struct WaylandNativeHandle{};
+
+#endif
 
 enum class NativeHandleType : char
 {
--- a/include/LLGL/Platform/Linux/LinuxNativeHandle.h
+++ b/include/LLGL/Platform/Linux/LinuxNativeHandle.h
@@ -14,7 +14,7 @@
 
 #include <LLGL/Deprecated.h>
 
-#if LLGL_LINUX_ENABLE_WAYLAND
+#if LLGL_EXPOSE_WAYLAND || LLGL_LINUX_ENABLE_WAYLAND
 #include <wayland-client.h>
 #endif
 
@@ -57,7 +57,7 @@ struct WaylandNativeHandle
 
 #endif
 
-enum class NativeHandleType : char
+enum class NativeType : char
 {
     X11 = 0,
     Wayland = 1
@@ -66,6 +66,8 @@ enum class NativeHandleType : char
 //! Linux native handle structure.
 struct NativeHandle
 {
+    NativeType type;
+
     union {
         struct {
             //! X11 display connection.
@@ -92,8 +94,6 @@ struct NativeHandle
         X11NativeHandle x11;
         WaylandNativeHandle wayland;
     };
-
-    NativeHandleType type = NativeHandleType::X11;
 };
 
 
--- a/include/LLGL/Platform/Linux/LinuxNativeHandle.h
+++ b/include/LLGL/Platform/Linux/LinuxNativeHandle.h
@@ -63,25 +63,27 @@ enum class NativeHandleType : char
 struct NativeHandle
 {
     union {
-        //! X11 display connection.
-        LLGL_DEPRECATED("Deprecated since 0.04b; Use x11.display instead.")
-        ::Display*      display;
+        struct {
+            //! X11 display connection.
+            LLGL_DEPRECATED("Deprecated since 0.04b; Use x11.display instead.")
+            ::Display*      display;
 
-        //! X11 window object.
-        LLGL_DEPRECATED("Deprecated since 0.04b; Use x11.window instead.")
-        ::Window        window;
+            //! X11 window object.
+            LLGL_DEPRECATED("Deprecated since 0.04b; Use x11.window instead.")
+            ::Window        window;
 
-        //! X11 visual information.
-        LLGL_DEPRECATED("Deprecated since 0.04b; Use x11.visual instead.")
-        ::XVisualInfo*  visual;
+            //! X11 visual information.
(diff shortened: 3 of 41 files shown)


CMakeLists.txt
```diff
@@ -176,6 +176,10 @@ if(WIN32)
     option(LLGL_BUILD_RENDERER_DIRECT3D12 "Include Direct3D12 renderer project" ON)
 endif()
 
+if(LLGL_TARGET_PLATFORM STREQUAL "Linux")
```
LukasBanana, line 179: `LLGL_TARGET_PLATFORM` in the CMake script is only intended as output for the summary, not for conditionals. You can  use the CMake variable `UNIX` for this condition instead.

include/LLGL/Backend/OpenGL/Linux/LinuxNativeHandle.h
```diff
@@ -26,7 +31,13 @@ namespace OpenGL
 */
 struct RenderSystemNativeHandle
 {
-    GLXContext context;
+    union
+    {
+        GLXContext glx;
```
LukasBanana, line 36: This is a breaking change so we'd have to keep the old name and potentially deprecate it unless you just add the new one, e.g. `contextEGL`.
Besides that, does Wayland only work with EGL?

include/LLGL/Platform/Linux/LinuxNativeHandle.h
```diff
@@ -36,6 +37,37 @@ struct NativeHandle
...
+{
+    X11 = 0,
+    Wayland = 1
+};
+
+//! Linux native handle structure.
+struct NativeHandle
+{
+    union {
+        X11NativeHandle x11;
```
LukasBanana, line 64: We have the same type of deprecation problem here. How about this, keep this union as is, but also keep the old members inside this `NativeHandle` struct. When an X11 window is used, we initialize both the old members and the `x11` field and when Wayland is used we initialize only the `wayland` field and reset the rest to zero. So something like this:
```cpp
struct NativeHandle
{
  LLGL_DEPRECATED("Deprecated since 0.04b; Use x11.display instead.")
  ::Display* display;

  ...

  union
  {
    X11NativeHandle x11;
    WaylandNativeHandle wayland;
  }
};
```

include/LLGL/Surface.h
```diff
@@ -88,6 +88,15 @@ class LLGL_EXPORT Surface : public Interface
         LLGL_DEPRECATED("Surface::ResetPixelFormat is deprecated since 0.04b; No need to reset pixel format anymore!")
         virtual void ResetPixelFormat();
 
+        /**
+        \brief Is the surface is using Wayland
+        \remarks Only works on Linux
+        \return True if the surface is using Wayland
+        */
+        virtual bool IsWayland() const {
```
LukasBanana, line 96: I think this new interface function is redundant if we already have this information through `NativeHandle::type`. You just call `GetNativeHandle()` and this field tells you what the surface type is.

sources/Renderer/OpenGL/Platform/Linux/LinuxGLSwapChainContext.cpp
```diff
@@ -22,58 +26,123 @@ namespace LLGL
 
 std::unique_ptr<GLSwapChainContext> GLSwapChainContext::Create(GLContext& context, Surface& surface)
 {
-    return MakeUnique<LinuxGLSwapChainContext>(static_cast<LinuxGLContext&>(context), surface);
+#ifdef LLGL_LINUX_ENABLE_WAYLAND
+    if (surface.IsWayland()) {
```
LukasBanana, line 30: Please use [Allman indentation](https://en.wikipedia.org/wiki/Indentation_style#Notable_styles) style in this project.

sources/Renderer/OpenGL/Platform/Linux/LinuxGLSwapChainContext.cpp
```diff
@@ -22,58 +26,123 @@ namespace LLGL
...
+{
+    // dummy
+}
+
+bool LinuxX11GLSwapChainContext::MakeCurrentUnchecked()
+{
+    return glXMakeCurrent(dpy_, wnd_, glc_);       
+}
+
+bool LinuxX11GLSwapChainContext::Destroy() {
```
LukasBanana, line 81: The idea behind all the `MakeContextCurrent` functions being static members is so that you can also pass a null pointer to handle the case of removing the current context from the internal GL state. This is intended to be independent of any LLGL owned context to be alive, so this refactoring of `MakeCurrentUnchecked()`/`Destroy()` being non-static member functions breaks this concept. In what way is this refactor necessary for Wayland support?

LukasBanana (changes requested):
Overall good job on the proposal, I'm excited to get Wayland support into LLGL as I haven't had the time to look into this API myself.

A few things so far:
1. I like how you split up the GLSwapChainContext class into two, one for X11 and one for Wayland.
2. I don't think the refactoring of the how the `MakeCurrent*Context` functions operate is necessary for Wayland support. I saw you introduced `MakeCurrentUnchecked()` and `Destroy()` as member functions but this breaks the initial idea of having them one single function that is independent of the active object, so you can withdraw any active GL context regardless of the previous object still being alive. And what if LLGL is used with a custom GL context in which case LLGL needs to be able to make those changes without even having an instance of `GLSwapChainContext` allocated yet.
3. Also please keep backwards compatibility in mind. I'm trying to make breaking changes a very rare case now, so I think we should deprecate old fields from the public interface instead of just replacing them with new ones.

st0rmbtw:
> 2\. I don't think the refactoring of the how the `MakeCurrent*Context` functions operate is necessary for Wayland support. I saw you introduced `MakeCurrentUnchecked()` and `Destroy()` as member functions but this breaks the initial idea of having them one single function that is independent of the active object, so you can withdraw any active GL context regardless of the previous object still being alive. And what if LLGL is used with a custom GL context in which case LLGL needs to be able to make those changes without even having an instance of `GLSwapChainContext` allocated yet.

There is a problem with that function:
```cpp
 bool GLSwapChainContext::MakeCurrentUnchecked(GLSwapChainContext* context) 
 { 
     return LinuxGLSwapChainContext::MakeCurrentGLXContext(static_cast<LinuxGLSwapChainContext*>(context)); 
 } 
```

include/LLGL/Backend/OpenGL/Linux/LinuxNativeHandle.h
st0rmbtw: > Besides that, does Wayland only work with EGL?

AFAIK, yes.

LukasBanana:
I'm sure we can use `glXGetCurrentContext()` for this to determine whether a GLX context is the current one, I don't think (not 100% sure, though) this would return a non-null pointer if a GL context was previously made current with EGL. Then there's also `eglGetCurrentContext()` to retrieve the active EGL context. The way you have it set up right now is what I had much earlier when I started with LLGL, but using an LLGL-instantiated object to unset a GL context makes not much sense so I think we should keep this decoupled.

include/LLGL/Platform/Linux/LinuxNativeHandle.h
```diff
@@ -12,13 +12,16 @@
 #include <X11/Xlib.h>
 #include <X11/Xutil.h>
 
+#include <LLGL/Deprecated.h>
+
+#ifdef LLGL_LINUX_ENABLE_WAYLAND
+#include <wayland-client.h>
```
LukasBanana, line 18: In GLFW for instance, they use a macro to expose Wayland that's separate from the build configuration. I think we should do something similar instead of relying on build configuration macros. Right now this macro would basically have to act as such an "exposure" macro so the client programmer has to include this header like this:
```cpp
#define LLGL_LINUX_ENABLE_WAYLAND
#include <LLGL/Platform/NativeHandle.h>
```
Using a separate macro would still require this pattern but at least it makes it clear that it's only meant for exposing it on the client level and independent of how LLGL was built. It would actually be nice to just have the client include either one of two headers, e.g. `<LLGL/Platform/NativeHandleX11.h>`/`NativeHandleWayland.h` but that makes it harder to get this to work when LLGL is built with support for both systems, so let's not worry about that at the moment (just thinking out loud here).

I'd say while Wayland support is experimental, we keep the X11 requirement as-is but only expose this in the public header with a macro like `LLGL_EXPOSE_WAYLAND` but skip this for X11 (i.e. no `LLGL_EXPOSE_X11` macro for now). It should of course automatically be enabled if LLGL is built with Wayland support, so maybe do something like this:
```cpp
#if LLGL_EXPOSE_WAYLAND || LLGL_LINUX_ENABLE_WAYLAND
#   include <wayland-client.h>
#endif
```

include/LLGL/Platform/Linux/LinuxNativeHandle.h
```diff
@@ -36,6 +39,59 @@ struct NativeHandle
     int             screen;
 };
 
+#ifdef LLGL_LINUX_ENABLE_WAYLAND
```
LukasBanana, line 42: Some minor nitpick: I started to rely much more on `#if` instead of `#ifdef` in case those macros are defined as `0`. CMake automatically defines them to `1` when you don't specify a value, so this will still work.

LukasBanana (changes requested):
Were you planning on adding Wayland support to the Linux implementation of the `Window` interface in a separate PR? I don't see it in here, so right now we can't test your change easily, at least not contained within this project. It would be nice to have at least some rudimentary Wayland support in the initial PR, i.e. maybe without support for any of the `WindowFlags` entries and only very basic window event handling. But it would be great if we were at least able to create a window and render something into. Have you been able to test this with another third party windowing library like GLFW?

st0rmbtw:
> Were you planning on adding Wayland support to the Linux implementation of the `Window` interface in a separate PR?

Yes. I initially wanted to do that in this PR as well, but when I was implementing it I ran into a problem I didn't know how to fix. Wayland API is a pure mess. You can check the implementation here: https://github.com/st0rmbtw/LLGL/tree/linux-wayland-window

> Have you been able to test this with another third party windowing library like GLFW?

Yes, of course. I wouldn't make a PR without ensuring that it works :).
It works, but it segfaults on exit and I'm not sure if it's a bug in Wayland itself or it's my fault.

sources/Renderer/OpenGL/Platform/MacOS/MacOSGLSwapChainContext.mm
```diff
@@ -30,7 +30,6 @@
     return MacOSGLSwapChainContext::MakeCurrentNSGLContext(static_cast<MacOSGLSwapChainContext*>(context));
 }
 
-
```
LukasBanana, line 33: There are some probably unintentional changes remaining in these files. You can simply revert them by checking out the source from the upstream repo.

sources/Renderer/Vulkan/Ext/VKExtensionLoader.cpp
```diff
@@ -177,11 +187,14 @@ bool VKLoadInstanceExtensions(VkInstance instance, const ArrayView<const char*>&
 
     /* Load platform specific extensions */
     #if defined LLGL_OS_WIN32
-    LOAD_VKEXT( KHR_win32_surface );
+        LOAD_VKEXT( KHR_win32_surface );
     #elif defined LLGL_OS_LINUX
-    LOAD_VKEXT( KHR_xlib_surface );
+        LOAD_VKEXT( KHR_xlib_surface );
+        #ifdef LLGL_LINUX_ENABLE_WAYLAND
```
LukasBanana, line 193: Remember to also update the remaining uses of `#ifdef` for this macro.

sources/Renderer/Vulkan/VKSwapChain.cpp
```diff
@@ -370,16 +374,34 @@ void VKSwapChain::CreateGpuSurface()
...
+        createInfo.window   = nativeHandle.x11.window;
+
+        VkResult result = vkCreateXlibSurfaceKHR(instance_, &createInfo, nullptr, surface_.ReleaseAndGetAddressOf());
+        VKThrowIfFailed(result, "failed to create Xlib surface for Vulkan swap-chain");
     }
-    VkResult result = vkCreateXlibSurfaceKHR(instance_, &createInfo, nullptr, surface_.ReleaseAndGetAddressOf());
-    VKThrowIfFailed(result, "failed to create Xlib surface for Vulkan swap-chain");
+    #ifdef LLGL_LINUX_ENABLE_WAYLAND
+    // Create Wayland surface
+    else if (nativeHandle.type == NativeHandleType::Wayland)
```
LukasBanana, line 392: Since we currently always support X11, can we only check for Wayland and use X11 in the else branch or at least have an else branch in case `nativeHandle` is corrupted. That else branch can simply contain `LLGL_TRAP("Invalid native handle type");` or something like that.

sources/Renderer/Vulkan/Vulkan.h
```diff
@@ -15,6 +15,7 @@
 #   define VK_USE_PLATFORM_WIN32_KHR
 #elif defined LLGL_OS_LINUX
 #   define VK_USE_PLATFORM_XLIB_KHR
+#   define VK_USE_PLATFORM_WAYLAND_KHR
```
LukasBanana, line 18: Should this only be defined if `LLGL_LINUX_ENABLE_WAYLAND` is enabled?

sources/Renderer/OpenGL/Profile/GLCore/OpenGLCore.h
```diff
@@ -21,6 +21,9 @@
 #   include <GL/gl.h>
 #   include <GL/glext.h>
 #   include <GL/glx.h>
+#   include <EGL/egl.h>
```
LukasBanana, line 24: This probably also needs to be wrapped in `LLGL_LINUX_ENABLE_WAYLAND`. Not everyone has EGL installed.

sources/Renderer/OpenGL/Platform/Linux/LinuxGLContext.cpp
```diff
@@ -205,57 +280,217 @@ bool LinuxGLContext::SetSwapInterval(int interval)
...
         return (glXSwapIntervalSGI(interval) == 0);
     }
+#ifdef LLGL_LINUX_ENABLE_WAYLAND
+    }
+#endif
 
     return false;
 }
 
+#ifdef LLGL_LINUX_ENABLE_WAYLAND
```
LukasBanana: Can we split this into separate source files? One for X11 and one for Wayland, because this is a pretty large blend between X11 and Wayland code in the same source file. You don't have to split it into separate classes but do something similar to what I did in _LinuxGLSwapChain.cpp_ where I only moved the Linux specific code for the _GLSwapChain_ class into its own file.

st0rmbtw:
@LukasBanana I've split `GLContext` into separate classes. Please, check if everything is done correctly

sources/Renderer/OpenGL/Platform/Linux/LinuxGLContextX11.cpp
```diff
@@ -0,0 +1,369 @@
...
+    }
+    else
+    {
+        CreateGLXContext(pixelFormat, profile, nativeWindowHandle, sharedContext);
+    }
+}
+
+LinuxGLContextX11::~LinuxGLContextX11()
+{
+    if (!isProxyGLC_) {
```
LukasBanana, line 68: It looks like this implementation has not been copied over into a new class 1:1. Besides the different formatting style, `isProxyGLC_` is not set to true in the upper branch that calls `CreateProxyGLXContext()` anymore. Ideally, this file should be almost identical to what the old `LinuxGLContext.cpp looked like, except for the `GLContext::Create()` implementation of course. Can you please revise this file by simply copying the old LinuxGLContext.cpp file and remove the parts that have to stay in the new version of that file? FWIW: I assume git might even recognize this file as being moved and modified in that case.

sources/Renderer/OpenGL/Platform/Linux/LinuxGLContextWayland.h
```diff
@@ -0,0 +1,109 @@
...
+#include "../../OpenGL.h"
+#include <LLGL/RendererConfiguration.h>
+#include <LLGL/Platform/NativeHandle.h>
+#include <X11/Xlib.h>
+
+
+namespace LLGL
+{
+
+enum class LinuxGLAPIType : char {
```
LukasBanana, line 25: Is this enum still needed? I presume this was from your initial implementation before you split up the classes. Your `IsWayland()` function solves this more elegantly IMO.

sources/Renderer/OpenGL/Platform/Linux/LinuxGLContextWayland.cpp
```diff
@@ -0,0 +1,229 @@
...
+    {
+        context_ = CreateEGLContextCoreProfile(glcShared, profile.majorVersion, profile.minorVersion, pixelFormat.depthBits, pixelFormat.stencilBits, &config_);
+    }
+
+    if (context_)
+    {
+        if (eglMakeCurrent(display, EGL_NO_SURFACE, EGL_NO_SURFACE, intermediateGlc) != True)
+            Log::Errorf("eglMakeCurrent failed on EGL core profile\n");
+
+        /* Valid core profile created, so we can delete the intermediate GLX context */
```
LukasBanana, line 51: Some comments here mention GLX but this should probably be EGL. Does the rest of those comments still apply or was this just copied from the X11 version of this class?

sources/Renderer/OpenGL/Platform/Linux/LinuxGLSwapChainContext.cpp
```diff
@@ -22,61 +23,151 @@ namespace LLGL
...
+
+#if LLGL_LINUX_ENABLE_WAYLAND
+
+#include "LinuxGLContextWayland.h"
+
+#include <wayland-egl-core.h>
+
+namespace LLGL {
+
+LinuxWaylandGLSwapChainContext::LinuxWaylandGLSwapChainContext(LinuxGLContextWayland& context, Surface& surface) :
```
LukasBanana, line 115: I'm glad to see you already split up `LinuxGLSwapChainContext` into two separate classes but can we also split the file the same way you did with _LinuxGLContext.cpp_? And can we use the same name pattern you used for `LinuxGLContext`, i.e. use `LinuxGLSwapChainContextWayland`/`X11`?

sources/Renderer/OpenGL/GLSwapChain.cpp
```diff
@@ -138,17 +150,12 @@ bool GLSwapChain::SetVsyncInterval(std::uint32_t vsyncInterval)
     return SetSwapInterval(static_cast<int>(vsyncInterval));
 }
 
-bool GLSwapChain::MakeCurrent(GLSwapChain* swapChain)
+bool GLSwapChain::MakeCurrent(GLSwapChain& swapChain)
```
LukasBanana, line 153: What's the purpose of this change? I see this function was only used with a reference at one location so far, but it was meant to have the same interface as the other MakeCurrent functions that also allow to remove the active swap-chain context.

LukasBanana (changes requested):
I think you getting into the right direction. We should apply the same separation of files in _LinuxGLSwapChainContext.cpp/.h_ as you did with _LinuxGLContext.cpp/.h_ to separate X11 and Wayland code. There's a few more cleanup to be done that I commented on, but I like where this is going. I'm fine with doing some of the refinements myself once we get this into the main stream.

LukasBanana:
> X Error of failed request:  BadMatch (invalid parameter attributes)

This looks like unsupported parameters, it's always a pain to interpret this error because it doesn't provide a lot of details. This might be due to a wrong `XVisualInfo` that was selected in `LinuxGLContext::ChooseVisual()` or wrong framebuffer attributes rather.
Can you try running the examples with `-ms=0` to disable multisampling? I am running this on my Ubuntu 24 VM in VirtualBox and this works with and without multisampling. Even though GL support has always been tricky inside a VM, mine even supports OpenGL 4.1 Core Profile. I had barely GL 2.1 support in the past when I started working on Linux support in LLGL :-/

> Looks like that error is unrelated to the changes made in this PR. I get the same error on the master branch as well.

There's also this error from only one of the Linux CI runs:
```
in 'Create': Native handle type is Wayland but LLGL was built without LLGL_LINUX_ENABLE_WAYLAND
```
It's not clear to me why this only shows up in the _Full/Shared/Release_ configuration, but it seems to happen during the _OffscreenC99_ test.

sources/Renderer/OpenGL/GLSwapChain.cpp
st0rmbtw: That change was made when I introduced the `Destroy` method to unset GL context.

sources/Renderer/OpenGL/Platform/Linux/LinuxGLContextWayland.cpp
st0rmbtw: > was this just copied from the X11 version of this class?

This. Sorry, I completely forgot about that.

st0rmbtw:
> Can you try running the examples with `-ms=0` to disable multisampling?

Disabling multisampling didn't help :(


> There's also this error from only one of the Linux CI runs:
> 
> ```
> in 'Create': Native handle type is Wayland but LLGL was built without LLGL_LINUX_ENABLE_WAYLAND
> ```

That's weird...

st0rmbtw:
Looks like https://github.com/LukasBanana/LLGL/pull/193/commits/965f1b1702d16b2b116431ee95269e67893ef0a4 fixed the issue with the `OffscreenC99` example. Though I don't understand why it didn't work, because the `X11` variant of the `NativeHandleType` enum is zero, so it should have worked. Maybe stack variable are not zero initialized by default in Release mode?

include/LLGL/Backend/OpenGL/Linux/LinuxNativeHandle.h
```diff
@@ -9,15 +9,24 @@
...
+#endif
+
 
 namespace LLGL
 {
 
 namespace OpenGL
 {
 
+enum class RenderSystemNativeHandleType : char {
```
LukasBanana, line 26: Minor nitpick: can we name this `RenderSystemNativeType` instead? This would match the pattern compared to `LLGL::Direct3D12::ResourceNativeHandle`/ResourceNativeType` and shorten this long identifier a little bit (as opposed to `LLGL::OpenGL::RenderSystemNativeHandleType::EGL`).

include/LLGL/Platform/Linux/LinuxNativeHandle.h
```diff
@@ -36,6 +39,63 @@ struct NativeHandle
...
+
+struct WaylandNativeHandle
+{
+    void* window;
+    void* display;
+};
+
+#endif
+
+enum class NativeHandleType : char
```
LukasBanana, line 60: Just as mentioned above with `RenderSystemNativeHandleType`, if we name it `RenderSystemNativeType` we should also rename this one to `NativeType`. I wanted to rename the "NativeHandle" to "SurfaceNativeHandle" at some point to clearly differentiate it from the render system native handles, but we might also just keep it as-is.

include/LLGL/Backend/OpenGL/Linux/LinuxNativeHandle.h
```diff
@@ -26,7 +35,19 @@ namespace OpenGL
...
+        GLXContext context;
+        GLXContext glx;
+        #if LLGL_EXPOSE_WAYLAND || LLGL_LINUX_ENABLE_WAYLAND
+        EGLContext egl;
+        #else
+        void*      egl;
+        #endif
+    };
+
+    RenderSystemNativeHandleType type;
```
LukasBanana, line 50: Can you make `type` the first field in this struct? `Direct3D12::ResourceNativeHandle` does the same. So the first bytes read are always identifying how to interpret the rest of the union.

sources/Renderer/OpenGL/Platform/GLSwapChainContext.h
```diff
@@ -52,7 +52,7 @@ class GLSwapChainContext
         // Creates a platform specific GLSwapChainContext instance.
         static std::unique_ptr<GLSwapChainContext> Create(GLContext& context, Surface& surface);
 
-        // Makes the specified swap-chain context link current. If null, no context is current.
+        // Primary function to make the specified swap-chain context link current.
```
LukasBanana, line 55: Is this comment change still valid? I guess it doesn't make a difference, but this doesn't mention the null pointer anymore. Same with the comment below. If this file doesn't need any changes, it also makes the PR a bit easier to navigate.

sources/Renderer/OpenGL/Platform/Android/AndroidGLSwapChainContext.cpp
```diff
@@ -63,7 +62,7 @@ void AndroidGLSwapChainContext::CanvasEventListener::OnInit(Canvas& sender)
 {
     /* Re-initialize the shared EGLSurface when the ANativeWindow is re-initialized */
     context_->InitEGLSurface(sender);
-    GLSwapChainContext::MakeCurrent(context_);
+    GLSwapChainContext::MakeCurrent(reinterpret_cast<GLSwapChainContext*>(context_));
```
LukasBanana, line 65: Do the Android source files need any changes? Might also be a left over from your original refactoring of the interface.

sources/Renderer/OpenGL/Platform/GLSwapChainContext.cpp
```diff
@@ -25,7 +25,7 @@ bool GLSwapChainContext::MakeCurrent(GLSwapChainContext* context)
     bool result = true;
     if (g_currentSwapChainContext != context)
     {
-        result = GLSwapChainContext::MakeCurrentUnchecked(context);
+        result = MakeCurrentUnchecked(context);
```
LukasBanana, line 28: This file probably also doesn't need any changes anymore.

sources/Renderer/OpenGL/Platform/Linux/LinuxGLContextWayland.cpp
```diff
@@ -0,0 +1,229 @@
...
+    eglDestroyContext(display_, context_);
+}
+
+void LinuxGLContextWayland::CreateProxyEGLContext(
+    const GLPixelFormat&                    pixelFormat,
+    const NativeHandle&                     nativeWindowHandle,
+    const OpenGL::RenderSystemNativeHandle& nativeContextHandle)
+{
+    LLGL_ASSERT(nativeWindowHandle.type == NativeHandleType::Wayland);
+    LLGL_TRAP("TODO");
```
LukasBanana, line 163: You can use `LLGL_TRAP_NOT_IMPLEMENTED("Wayland proxy EGL context")` for this.

sources/Renderer/OpenGL/Platform/Linux/LinuxGLContextWayland.h
```diff
@@ -0,0 +1,104 @@
+/*
+ * LinuxGLContext.h
+ *
+ * Copyright (c) 2025 Lukas Hermanns. All rights reserved.
+ * Licensed under the terms of the BSD 3-Clause license (see LICENSE.txt).
+ */
+
+
+#if LLGL_LINUX_ENABLE_WAYLAND
```
LukasBanana, line 9: The header guard should always come first. Ideally we should exclude the entire file in the CMake script when Wayland support is disabled, but we can leave it to this preprocessor directive for now.

sources/Renderer/OpenGL/Platform/Linux/LinuxGLContextX11.h
```diff
@@ -0,0 +1,100 @@
+/*
+ * LinuxGLContext.h
+ *
+ * Copyright (c) 2025 Lukas Hermanns. All rights reserved.
```
LukasBanana, line 4: Another case of changed copyright notice. Please keep those unchanged except for the filename at the top.

sources/Renderer/OpenGL/Platform/Linux/LinuxGLContextX11.cpp
```diff
@@ -0,0 +1,362 @@
+/*
+ * LinuxGLContextX11.cpp
+ *
+ * Copyright (c) 2025 Lukas Hermanns. All rights reserved.
```
LukasBanana, line 4: And another one with changed copyright notice.

sources/Renderer/OpenGL/Platform/Linux/LinuxGLSwapChainContext.cpp
```diff
@@ -1,14 +1,18 @@
 /*
  * LinuxGLSwapChainContext.cpp
  *
- * Copyright (c) 2015 Lukas Hermanns. All rights reserved.
+ * Copyright (c) 2025 Lukas Hermanns. All rights reserved.
```
LukasBanana, line 4: Same as above.

LukasBanana (changes requested):
Okay this should be the last request for changes.

I left a few comments for some cleanup and please copy the copyright notice from one of the original files as-is except for the filename. I noticed several `Copright (c) 2025` lines but I stopped updating them year after year, so these should not be changed also not for new files.

There's a few more things with formatting and naming that I can take care of after I merge this, but I added a few comments about that, too.

And some files should not be part of this PR, like the Android source files for instance and some others that might still be left overs from your original refactoring of the current-context changes.

Overall, good job on the introduction of Wayland support 👍 I'm impressed what you came up with so far, especially under the consideration that you just started working on Linux :)

LukasBanana:
> Looks like https://github.com/LukasBanana/LLGL/commit/965f1b1702d16b2b116431ee95269e67893ef0a4 fixed the issue with the OffscreenC99 example.

Oh I overlooked that change. We can't keep that in, though, because these structs need to be POD structs, so we can't have default initializers for them.
I think the bug is that you also have to output the native type for `X11` in `LinuxWindow::GetNativeHandle()` and perhaps also zero-initialize the struct in `GLContext::Create()`:
```cpp
    LLGL::NativeHandle nativeHandle;
    surface.GetNativeHandle(&nativeHandle, sizeof(nativeHandle));
```
This will have an uninitialized value for the type. So just add `= {}` to zero-initialize it and also have `LinuxWindow` output what type of a Window it is. Since we don't support Wayland for the windowing system yet, this will always be X11.

st0rmbtw:
@LukasBanana Suggestions have been applied :)

LukasBanana (approved):

LukasBanana:
Since this PR became a bit larger with some cleanup left for me, I wanted to merge this into a separate branch first, but ended up just merging it into the master branch with 8eab7f8.

Thanks again for your contribution, I'm excited to get a more broader Wayland support into LLGL with its default windowing system.

PR #197, opened 2025-05-17
Don't hardcode the search path for DLLs

st0rmbtw:


st0rmbtw:
@LukasBanana Should I do the same for `Android`?

LukasBanana:
Yeah Android is pretty much a copy of the Linux module for this class. Makes sense to do the same there.

st0rmbtw:
Yeah, I also noticed that CI is failing on Linux at the Testbed stage because it sets `tests/Testbed` as the working directory, but the DLLs are in the `build` directory.

LukasBanana:
> But how does Window understand where to look for DLLs?

There is a long list of paths Windows searches for DLLs which I remember also being referred to as "DLL Hell":
https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order#standard-search-order-for-unpackaged-apps

LukasBanana:
I think loading shared libraries from the working directory is rather discouraged on Linux. The recommended way appears to be either use the RPATH or the search path provided by the system:
https://clarkkromenaker.com/post/library-dynamic-loading-linux/

This has likely security reasons. Only on Windows, this is all over the place.

st0rmbtw:
@LukasBanana I implemented fallbacking to the program path if DLLs were not found in the current directory. Please, check it out.

st0rmbtw:
> I think loading shared libraries from the working directory is rather discouraged on Linux. The recommended way appears to be either use the RPATH or the search path provided by the system:
> https://clarkkromenaker.com/post/library-dynamic-loading-linux/

Ohh, should I revert the commit I just made?

LukasBanana:
Yeah sorry I made you revise this before checking. But I think we should use the RPATH instead of trying to emulate the behavior from Windows.

st0rmbtw:
> But I think we should use the RPATH instead of trying to emulate the behavior from Windows.

I've just googled up what RPATH is and found out that [RPATH is deprecated and should be avoided. RUNPATH is preferred because it can be overridden by LD_LIBRARY_PATH](https://stackoverflow.com/a/54723461).

LukasBanana:
I see. This would still be provided in the executable, i.e. the respective project, not _libLLGL.so_, right?

st0rmbtw:
> I see. This would still be provided in the executable, i.e. the respective project, not _libLLGL.so_, right?

Uhh, I don't know. There is a [BUILD_RPATH](https://cmake.org/cmake/help/latest/prop_tgt/BUILD_RPATH.html) in CMake, but I'm not sure which one does it set `RPATH` or `RUNPATH`.

st0rmbtw:
Let me clarify, we should not search for DLLs by ourselves on Linux but let the OS do it instead, right? So for example we just do
```cpp
dlopen("libLLGL_Vulkan.so")
```

and that's it? So we don't interact with the `RUNPATH` or `RPATH` things directly?

LukasBanana:
I think that's correct. We should probably get the program path into the `RUNPATH` instead of adding that path at runtime. This whole `GetProgramPath()` was always more of a workaround.

CMakeLists.txt
```diff
@@ -445,6 +445,7 @@ function(add_llgl_example_project PROJECT_NAME LINKER_LANG SRC_FILES LIB_FILES)
         ${PROJECT_NAME} PROPERTIES
             DEBUG_POSTFIX               "D"
             RUNTIME_OUTPUT_DIRECTORY    "${LLGL_OUTPUT_DIR}"
+            BUILD_RPATH                 "${LLGL_OUTPUT_DIR}"
```
LukasBanana, line 448: This should be `\$ORIGIN` instead of `${LLGL_OUTPUT_DIR}` so that you can move the executable and still load it from the directory where the program is currently located. This way you're fixing this old bandaid and still have the behavior unchanged.
I tested both solutions and this is what `objdump` shows when inspecting `libLLGL.so`:
```
$ objdump -x libLLGL.so |grep 'R.*PATH'
  RUNPATH              /home/lh/Development/Projects/LLGL/repo/build_linux/build:
$ objdump -x libLLGL.so |grep 'R.*PATH'
  RUNPATH              $ORIGIN:
```

LukasBanana:
Looks good. Thanks for improving this ❤️

PR #200, opened 2025-05-20
Wayland windows

st0rmbtw:


CMakeLists.txt
```diff
@@ -746,6 +745,31 @@ else()
     set_target_properties(LLGL PROPERTIES BUILD_RPATH "\$ORIGIN")
 endif()
 
+# Generate Wayland protocols
+if(LLGL_LINUX_ENABLE_WAYLAND)
+    file(MAKE_DIRECTORY ${PROJECT_SOURCE_DIR}/sources/Platform/Linux/protocols)
+
+    add_custom_command(
+        COMMAND wayland-scanner private-code < /usr/share/wayland-protocols/stable/xdg-shell/xdg-shell.xml > xdg-shell-protocol.c
```
LukasBanana, line 753: I like the idea of having this generated on the fly instead of having these third party source files being part of the repository due to licensing. However, I'm having some trouble building this when running `$ ./BuildLinux.sh --wayland`:
```
[  0%] Generating /home/lh/Development/Projects/LLGL/repo/sources/Platform/Linux/protocols/xdg-shell-protocol.c
/bin/sh: 1: cannot open /usr/share/wayland-protocols/stable/xdg-shell/xdg-shell.xml: No such file
gmake[2]: *** [CMakeFiles/LLGL.dir/build.make:78: /home/lh/Development/Projects/LLGL/repo/sources/Platform/Linux/protocols/xdg-shell-protocol.c] Error 2
gmake[1]: *** [CMakeFiles/Makefile2:245: CMakeFiles/LLGL.dir/all] Error 2
gmake: *** [Makefile:136: all] Error 2
```
The CI scripts for Linux also won't build with Wayland unless you enable this CMake option in _ci_linux.yml_.

sources/Platform/Linux/LinuxWindowWayland.cpp
```diff
@@ -0,0 +1,948 @@
...
+    if (!state_.xdg_toplevel)
+        LLGL_TRAP("Failed to get XDG toplevel");
+
+    xdg_toplevel_add_listener(state_.xdg_toplevel, &xdg_toplevel_listener, this);
+
+    xdg_toplevel_set_title(state_.xdg_toplevel, state_.title);
+    wl_surface_commit(state_.wl_surface);
+}
+
+bool LinuxWindowWayland::ProcessEvents()
```
LukasBanana, line 798: The `ProcessEvents()` function must remain a static function and it's meant to process the events for all surfaces (Canvas or Window). When Wayland is enabled, the `Window::Create()` function should automatically create Wayland windows instead of X11 windows; maybe with some runtime condition, i.e. even if it's enabled programmatically, we still need to check if the environment even supports Wayland and fall back to X11 otherwise.

Ideally, you would only create either one or the other, not both. The same condition needs to be used whether to process all X11 or Wayland events. If it's easier to handle this with a member function to access the Wayland-state specifically for each window instance, you can add a separate function that is public, but only internally:
```cpp
// .h
class LinuxWaylandWindow {
  public:
    // Can be public because LinuxWaylandWindow is not publicly accessible
    // and this function is only needed by the public function Surface::ProcessEvents().
    void ProcessWaylandEvents();
};

// .cpp
void Surface::ProcessEvents()
{
  #if LLGL_LINUX_ENABLE_WAYLAND
  if ( IsWaylandSupportedAtRuntime ? ) {
    foreach (waylandWindow) {
      waylandWindow->ProcessWaylandEvents();
    }
  }
  else
  #endif
  {
    foreach (x11Window) {
      x11Window->ProcessX11Events();
    }
  }
}
```

LukasBanana (changes requested):

sources/Platform/Linux/LinuxWindowWayland.cpp
st0rmbtw: > even if it's enabled programmatically, we still need to check if the environment even supports Wayland and fall back to X11 otherwise.

I think if the environment doesn't support Wayland then the app won't even build because of missing Wayland headers.

sources/Renderer/OpenGL/Platform/Linux/LinuxGLContextWayland.cpp
```diff
@@ -52,9 +63,20 @@ void LinuxGLContextWayland::CreateEGLContext(
         context_ = CreateEGLContextCoreProfile(glcShared, profile.majorVersion, profile.minorVersion, pixelFormat.depthBits, pixelFormat.stencilBits, &config_);
     }
 
+    if (sharedContext != nullptr)
```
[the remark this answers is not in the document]
st0rmbtw: > What do the examples look like on your machine?

It looks like this:
image

Something is really off with the alpha channel. And by the way it didn't crash at `eglMakeCurrent` on my machine.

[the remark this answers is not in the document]
LukasBanana: Have you had a chance to take another stab at this? I might have more time next week to test this again in my Linux VM.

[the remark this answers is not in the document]
st0rmbtw: No, I haven't. But I'm taking last exam on July 3rd, so after that I will finally have time to work on this.

LukasBanana:
I can see the window decorations. Great progress! But I still need to disable the previously mentioned creation of the EGL surface and secondary `eglMakeCurrent()` call. That might be a limitation of my VM not supporting any EGL configurations that can render both into windows and pbuffers:
image

The shared EGL surface was needed for Android because of this:
```cpp
// EGLSurface is shared because it is required when the initial EGLContext is made current.
// By that time, we won't have a user created surface but we have the native android_app window.
// So this surface is created with the first AndroidGLContext and is then shared with subsequently
// created AndroidGLSwapChainContext objects.
```

Besides all that, we still have to fix the front-facing issue, which might be due to a different coordinate system in Wayland compared to X11.

Interestingly, resizing the window is pretty slow, but that might also be an issue with my VM setup:

https://github.com/user-attachments/assets/3d5387f8-9824-4002-aae1-833033316d74

st0rmbtw:
@LukasBanana I fixed a bug when the window was transparent with the OpenGL backend on Wayland in 8ddc598af05fe7bb445b1c5d16fc8973c99c01c0. Now I don't have any issues with the OpenGL backend on Wayland on my PC. 
image

> Interestingly, resizing the window is pretty slow, but that might also be an issue with my VM setup

It's also slow for me, but not as much as shown in your video, I think it might be Wayland's issue, but I'm not sure.

> But I still need to disable the previously mentioned creation of the EGL surface and secondary `eglMakeCurrent()` call.

I don't have a such issue, might be a VM limitation indeed, but not sure.

CCobaltDev:
Looking forward to this

st0rmbtw:
Okay, I think this PR is ready to be undrafted. Some Unity checks are failing in the CI, but I don't think it's related to this PR.

st0rmbtw:
@LukasBanana  Okay, I think this PR is ready to be undrafted. Some Unity checks are failing in the CI, but I don't think it's related to this PR.

sources/Renderer/OpenGL/Platform/Linux/LinuxGLCore.cpp
```diff
@@ -0,0 +1,49 @@
+/*
+ * LinuxGLCore.cpp
+ *
+ * Copyright (c) 2015 Lukas Hermanns. All rights reserved.
+ * Licensed under the terms of the BSD 3-Clause license (see LICENSE.txt).
+ */
+
+#include "LinuxGLCore.h"
+#include "../../../../Core/MacroUtils.h"
+
```
LukasBanana, line 10: Please add the following under the _MacroUtils.h_ include directive to resolve the unity build issue on Linux:
```cpp
// Unity build on Linux fails to include this macro from MacroUtils.h,
// possibly due to this file being outside the LLGL_OpenGL backend.
#ifndef LLGL_CASE_TO_STR
#   define LLGL_CASE_TO_STR(VALUE) \
        case VALUE: return #VALUE
#endif
```
Maybe the unity build with CMake doesn't work correctly here because the header is outside the LLGL_OpenGL backend. A bit annoying but I'll try to generalize this once your PR is merged because it's a copy of the _AndroidGLCore_ files.

sources/Renderer/OpenGL/Platform/Linux/LinuxGLCore.h
```diff
@@ -0,0 +1,33 @@
+/*
+ * LinuxGLCore.h
+ *
+ * Copyright (c) 2015 Lukas Hermanns. All rights reserved.
+ * Licensed under the terms of the BSD 3-Clause license (see LICENSE.txt).
+ */
+
+#ifndef LLGL_ANDROID_GL_CORE_H
```
LukasBanana: Should be `LLGL_LINUX_GL_CORE_H`.

sources/Renderer/OpenGL/Platform/Linux/LinuxGLContextWayland.cpp
```diff
@@ -39,12 +40,34 @@ void LinuxGLContextWayland::CreateEGLContext(
...
+        sharedSurface_ = sharedContext->GetSharedEGLSurface();
+    }
+    else
+    {
+        sharedSurface_ = std::make_shared<LinuxSharedEGLSurface>(display_, config_, nullptr);
+    }
+
+    EGLSurface nativeSurface = sharedSurface_->GetEGLSurface();
+
+    if (eglMakeCurrent(display, nativeSurface, nativeSurface, intermediateGlc) != EGL_TRUE)
```
LukasBanana, line 70: For the intermediate GL context, do we need to make this context current with the native surface? This still fails in my VM but the following works for me:
```cpp
#if 0
    if (sharedContext != nullptr)
    {
        /* Share EGLSurface with shared context */
        sharedSurface_ = sharedContext->GetSharedEGLSurface();
    }
    else
    {
        sharedSurface_ = std::make_shared<LinuxSharedEGLSurface>(display_, config_, nullptr);
    }

    EGLSurface nativeSurface = sharedSurface_->GetEGLSurface();
#else
    EGLSurface nativeSurface = EGL_NO_SURFACE;
#endif
```
Then I also commented out the secondary GL context in the following conditional branch:
```cpp
    if (context_)
    {
//        if (eglMakeCurrent(display, nativeSurface, nativeSurface, context_) != EGL_TRUE)
```
I also had to disable the PBUFFER flags in `SelectConfig()`, but as previously mentoned, this might just be a limitation of my VM. Nonetheless, do we have to make the intermediate context current with a native surface as this only seems to prevent my VM from initializing the context successfully. Can you please test if this also works on your native Linux machine?

LukasBanana (changes requested):
Overall great job. I only left a few comments. I can't say much to the Wayland code unfortunately, but it seems to work fine even in my VM, except the poor performance, but not sure if that's also VM related as I don't know how much more resources Wayland requires.

sources/Renderer/OpenGL/Platform/Linux/LinuxGLContextWayland.cpp
st0rmbtw: With the changes you suggested above I get transparent window and some problem with framebuffer clearing: 

image

For some reason the window is opaque with the black background in the screenshot, but it's actually transparent. Might be some issue with the screenshot tool.

LukasBanana:
I think the static unity builds are still failing because we missed to add EGL to the linker dependencies in the OpenGL CMake script. We added it to the `find_components()` statement but not to `target_link_libraries()`. This is what `${OPENGL_LIBRARIES}` emits on my machine and EGL is missing here:
```
/usr/lib/x86_64-linux-gnu/libOpenGL.so;/usr/lib/x86_64-linux-gnu/libGLX.so;/usr/lib/x86_64-linux-gnu/libGLU.so
```
Can you make the following change to the [OpenGL/CMakeLists.txt:258](https://github.com/LukasBanana/LLGL/blob/1808c75b3eabd6f27e0da8a15cea47f4d050cf61/sources/Renderer/OpenGL/CMakeLists.txt#L258) script and see if it resolves the CI builds?
```
        if(LLGL_LINUX_ENABLE_WAYLAND)
            target_link_libraries(LLGL_OpenGL LLGL EGL ${OPENGL_LIBRARIES})
        else()
            target_link_libraries(LLGL_OpenGL LLGL ${OPENGL_LIBRARIES})
        endif()
```
Unfortunately, this doesn't reproduce on my Ubuntu VM, so I'm not sure if that's going to fix it, just a shot in the dark.

LukasBanana:
Yeah it would be nicer to exclude those files in CMake, but we can leave it like this for now. At least CI builds are all green now.

sources/Platform/Linux/LinuxDisplay.cpp
```diff
@@ -134,187 +88,73 @@ bool Display::IsCursorShown()
...
-        /*root_return*/     &rootWndReturn,
-        /*child_return*/    &childWndReturn,
-        /*root_x_return*/   &rootPosition.x,
-        /*root_y_return*/   &rootPosition.y,
-        /*win_x_return*/    &childPosition.x,
-        /*win_y_return*/    &childPosition.y,
-        /*mask_return*/     &mask
-    );
-    return rootPosition;
+    return g_primaryDisplay->GetCursorPosition();
```
LukasBanana, line 96: Oh I think this needs to be `GetCursorPositionInternal()`, otherwise this would be an infinite recursion.

LukasBanana (approved):
I'll update the one spot I commented on myself so we can finally get your PR. Thanks again for your awesome work, I'm excited to finally see proper Wayland support in LLGL!

PR #202, opened 2025-06-16
Allow to change vertex format

coco875:
make just a pr for https://github.com/LukasBanana/LLGL/discussions/201don't really know how safe it is.

### DIFF
--- a/include/LLGL/Buffer.h
+++ b/include/LLGL/Buffer.h
@@ -58,6 +58,15 @@ class LLGL_EXPORT Buffer : public Resource
         */
         virtual BufferDescriptor GetDesc() const = 0;
 
+        /**
+        \brief Sets the vertex attributes of this buffer.
+        \param[in] vertexAttribs Array view of vertex attributes to set.
+        \remarks This function is only used for vertex buffers and ignored if \c bindFlags does not contain the BindFlags::VertexBuffer bit.
+        \see BindFlags::VertexBuffer
+        \see VertexShaderAttributes::inputAttribs
+        */
+        virtual void SetVertexAttribs(const ArrayView<VertexAttribute>& vertexAttribs) {};
+
     protected:
 
         Buffer(long bindFlags);
--- a/sources/Renderer/Direct3D11/Buffer/D3D11Buffer.cpp
+++ b/sources/Renderer/Direct3D11/Buffer/D3D11Buffer.cpp
@@ -432,8 +432,8 @@ void D3D11Buffer::WriteWithSubresourceCopyWithCpuAccess(
 
 void D3D11Buffer::SetVertexAttribs(const ArrayView<VertexAttribute>& vertexAttribs)
 {
-    if (!desc.vertexAttribs.empty())
-        stride_ = desc.vertexAttribs.front().stride;
+    if (!vertexAttribs.empty())
+        stride_ = vertexAttribs.front().stride;
 }
 
 
--- a/sources/Renderer/Direct3D11/Buffer/D3D11Buffer.cpp
+++ b/sources/Renderer/Direct3D11/Buffer/D3D11Buffer.cpp
@@ -430,6 +430,12 @@ void D3D11Buffer::WriteWithSubresourceCopyWithCpuAccess(
     WriteWithStagingBuffer(context, stagingBuffer.Get(), data, dataSize, dstOffset);
 }
 
+void D3D11Buffer::SetVertexAttribs(const ArrayView<VertexAttribute>& vertexAttribs)
+{
+    if (!desc.vertexAttribs.empty())
+        stride_ = desc.vertexAttribs.front().stride;
+}
+
 
 } // /namespace LLGL
 
--- a/sources/Renderer/OpenGL/Buffer/GL3PlusSharedContextVertexArray.h
+++ b/sources/Renderer/OpenGL/Buffer/GL3PlusSharedContextVertexArray.h
@@ -40,6 +40,13 @@ class GL3PlusSharedContextVertexArray
         // Sets the debug label for all VAOs.
         void SetDebugName(const char* name);
 
+        void Reset()
+        {
+            attribs_.clear();
+            contextDependentVAOs_.clear();
+            debugName_.clear();
+        }
+
     private:
 
         struct GLContextVAO
--- a/sources/Renderer/OpenGL/Buffer/GL3PlusSharedContextVertexArray.h
+++ b/sources/Renderer/OpenGL/Buffer/GL3PlusSharedContextVertexArray.h
@@ -43,6 +43,11 @@ class GL3PlusSharedContextVertexArray
         void Reset()
         {
             attribs_.clear();
+            for (auto& contextVAO : contextDependentVAOs_)
+            {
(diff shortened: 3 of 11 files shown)


sources/Renderer/Direct3D11/Buffer/D3D11Buffer.cpp
```diff
@@ -430,6 +430,12 @@ void D3D11Buffer::WriteWithSubresourceCopyWithCpuAccess(
     WriteWithStagingBuffer(context, stagingBuffer.Get(), data, dataSize, dstOffset);
 }
 
+void D3D11Buffer::SetVertexAttribs(const ArrayView<VertexAttribute>& vertexAttribs)
+{
+    if (!desc.vertexAttribs.empty())
```
LukasBanana, line 435: This is the typo that causes the Windows CIS builds to fail: `desc.` must be removed here.

include/LLGL/Buffer.h
```diff
@@ -58,6 +58,15 @@ class LLGL_EXPORT Buffer : public Resource
...
         virtual BufferDescriptor GetDesc() const = 0;
 
+        /**
+        \brief Sets the vertex attributes of this buffer.
+        \param[in] vertexAttribs Array view of vertex attributes to set.
+        \remarks This function is only used for vertex buffers and ignored if \c bindFlags does not contain the BindFlags::VertexBuffer bit.
+        \see BindFlags::VertexBuffer
+        \see VertexShaderAttributes::inputAttribs
+        */
+        virtual void SetVertexAttribs(const ArrayView<VertexAttribute>& vertexAttribs) {};
```
LukasBanana, line 68: I assume this only had a default implementation for debugging purposes, but for your final PR, pelase make this a pure virtual function. This means you'll also have to provide an implementation in the DebugLayer (i.e. `DbgBuffer`), Null backend (i.e. `NullBuffer`), and Metal backend (i.e. `MTBuffer`).

LukasBanana (changes requested):
I haven't had the situation where I had to change the vertex format, although it would be nice to be able to use a single vertex buffer with multiple different formats just like you can do with the index buffer. This would allow to use one section of the buffer for one format and another section with another format, but I doubt this has any performance benefits because you'd still have to bind the buffer every time you want to use a different format as binding the buffer determines how the vertices are fetched from the buffer in the vertex-shader. Changing the format seems even less efficient because now you'd have to re-create internal structures and containers. For GL you'd have to drop all previously created VAOs (although GL might be the odd one out here).

Can you provide an example where this is beneficiary and where copying the section of the buffer into another buffer with a different vertex format is not viable?
In other words, have you considered using [`CommandBuffer::CopyBuffer`](https://github.com/LukasBanana/LLGL/blob/master/include/LLGL/CommandBuffer.h#L132)?

If you still think this is a critical feature, can you please also add a unit test to the Testbed project? I can assist you with once you shared an example use case.

coco875:
the use case it's because we use one shader but we make different version of it https://github.com/coco875/libultraship/blob/llgl-revert/src/graphic/Fast3D/shaders/default.shader.fs so we end up with a ton of format and most of this format are use to render so for one frame we can end up using/switching to use a lot of format so if we make a buffer for each format it fill vram if we just create a vertex buffer for each upload slow the render, so update the vertex buffer are faster.

coco875:
(it's specifically [here](https://github.com/coco875/libultraship/blob/llgl-revert/src/graphic/Fast3D/backends/gfx_llgl.cpp#L774) where I use/need it)

sources/Renderer/OpenGL/Buffer/GL3PlusSharedContextVertexArray.h
```diff
@@ -40,6 +40,18 @@ class GL3PlusSharedContextVertexArray
         // Sets the debug label for all VAOs.
         void SetDebugName(const char* name);
 
+        void Reset()
```
LukasBanana, line 43: This function won't work when there is more than one GL context because each VAO only exists in the respective context it was created in, so you'd have to make the context current before deleting the respective VAO (There's a comment in `GLVertexArrayObject::Release()` that mentions something about this issue). Doing so in this function by iterating over all contexts would be extremely slow and removes the benefit of changing the vertex format on demand.

We don't have to delete the VAOs just to update the vertex attributes, but we need a way to determine whether they have been invalidated. This could be done similar to how the debug label is updated, which is marked via `isObjectLabelDirty` for all contexts in `GL3PlusSharedContextVertexArray::SetDebugName()` and only updated when the VAO is requested the next time in `GL3PlusSharedContextVertexArray::GetVAOForCurrentContext()`. We could do the same for invalidating the format; The second if-branch in `GetVAOForCurrentContext()` looks like a good candidate. `BuildVertexLayout()` already checks if the VAO is valid, if not it generates a new one, otherwise it only updates the attributes. Now it gets a little complicated because re-using the old VAO means we now also have to keep track of how many attributes there were before updating it. `GLVertexArrayObject::BuildVertexLayout()` will need to disable old attributes if the new format has less attributes than the previous one. You can take a look at `GLStateManager::DisableVertexAttribArrays()`, which is apparently no longer used but previously served a similar purpose.

LukasBanana (changes requested):
I left a comment on your `Release()` function. Besides that, I looked at your use case in `Fast::GfxRenderingAPILLGL::DrawTriangles()` and I am wondering why this can't be done without updating the vertex format on demand. You are updating the vertex buffer every time in that function right before changing the vertex attributes and those attributes are stored per shader program:
```cpp
llgl_cmdBuffer->UpdateBuffer(*vertexBuffer, 0, buf_vbo, buf_vbo_len * sizeof(float));
vertexBuffer->SetVertexAttribs(mCurrentShaderProgram->vertexFormat.attributes);
```
Why don't you store a vertex buffer with the respective format per shader program and update that one? That would eliminate the need to update its format in ever draw call. You also conditionally re-create the buffer in that function if it needs to increase in size: `if (vertexBufferSize < buf_vbo_len)`. So whenever you have to update the format for a shader program you could just re-create that buffer since its content is updated in ever draw call nonetheless.

Perhaps I'm missing something but maybe we can avoid the trouble with the GL VAO updates.

coco875:
already try to make one vertex buffer per shader but there is some quantity of different shader and it accumulate triangle, so it's not rarer to have 500 triangle draw in one command. So I endup taking all my vram.

sources/Renderer/OpenGL/Buffer/GL3PlusSharedContextVertexArray.h
coco875: hmm can try but I am not so much confident on how to do it. I will not being able to continue for a while because of another project.

LukasBanana:
I resolved this by implementing a secondary `SetVertexBuffer(Buffer&, uint32_t, const VertexAttribute*)` function that allows to change the vertex format of the buffer that is bound to the input assembly. Feel free to re-open this PR or open a new ticket if this doesn't work for you. Thanks again for bringing this up. I was trying to work with your PR as a base but there were too many other things to refactor, so I hope you don't mind me replace this PR with 9ce6648.

PR #212, opened 2025-07-12
fix "unformSize" caculation error  

swq0553:
opengl backend
when there are uniform arrays in shader script，uniform.count will not be 1.
see
`GLImmediateCommandBuffer::SetUniforms(...)` and `GLDeferredCommandBuffer::SetUniforms(...)`


LukasBanana:
Thank you for this bug fix. It took me a moment to verify this fixes an issue with the "Uniforms" unit test.

For future reference, please provide a small example or code snippet that shows what issue this fixes. Even better would be to submit an addition to the Testbed. I changed the `lightVec[2]` uniform parameter in the *DynamicTriangleMesh.330core.frag* shader to an array and re-arranged the order of the uniform declarations in the PSO layout.

Issue #217, opened 2025-07-20
[Windows] Vulkan backend performance is disastrous with `RenderSystemFlags::DebugDevice`


dolphineye74:
Hey,

I finally managed to get the Vulkan backend to work although it still has the validation errors you're already aware of.

LIke with D3D12, I turned on Vulkan validation checks, but it causes a tremendous performance hit (basically 1000 msec per frame).

```cpp
LLGL::RenderSystemDescriptor renderSystemDesc;
renderSystemDesc.moduleName = "Vulkan";
renderSystemDesc.flags |= LLGL::RenderSystemFlags::DebugDevice;
renderSystem = LLGL::RenderSystem::Load(renderSystemDesc);
```

Commenting that line out just fixes the issue though.

Did I do anything wrong?

dolphineye74:
I narrowed down the issue to this validation feature (commenting that line out fixes the problem):

```cpp
VK_VALIDATION_FEATURE_ENABLE_GPU_ASSISTED_EXT
```

GPU information reported by LLGL:

```
Renderer: Vulkan 1.3.302
Device: AMD Radeon RX 6950 XT
Vendor:
Shading language: SPIR-V
```

dolphineye74:
Is there a way to potentially configure which validation features we want or not when we create the render system?

At this point, the app running at 1 FPS just for `VK_VALIDATION_FEATURE_ENABLE_GPU_ASSISTED_EXT` isn't ideal.

I still get the validation messages w/o `VK_VALIDATION_FEATURE_ENABLE_GPU_ASSISTED_EXT` and the app is running smoothly again.

dolphineye74:
A CMake option works just fine. I can possibly submit a PR with such a new option to save you up some time

LukasBanana:
You can do something similar as `LLGL_VK_ENABLE_SPIRV_REFLECT` in Renderer/Vulkan/CMakeLists.txt:18.

PR #218, opened 2025-07-22
[Vulkan] Add CMake option to disable GPU assisted validation

dolphineye74:
# Description
As reported in https://github.com/LukasBanana/LLGL/issues/217, combining Vulkan CPU & GPU assisted validation can cause terrible performance. On occasions, one might want to enable CPU validation only and enable GPU validation on demand.

A new `LLGL_VK_ENABLE_GPU_ASSISTED_VALIDATION` CMake option allows one to disable GPU validation if needed. By default the option was kept to `ON` for retrocompatibility purposes.

On Apple platforms, the option is not available, which forcefully disables GPU assisted validation. I took the liberty to do so given there are reports of [issues with GPU-VA with MoltenVK](https://github.com/KhronosGroup/MoltenVK/issues/2492). Let me know if you feel like this is a mistake.

In the C++ side, 

# Testing done
As I'm not sure if there is a way to craft a CI test case for this change, so I performed manual testing.

Test cases:
* Enabled `LLGL_VK_ENABLE_GPU_ASSISTED_VALIDATION` (default behavior) and verified the GPU assisted validation was turned on
* Disabled `LLGL_VK_ENABLE_GPU_ASSISTED_VALIDATION` and verified the GPU assisted validation was turned off

Tested platform:
* OS: Windows 11 Professional 24H2
* Backend: Vulkan backend
* GPU: AMD RX 6950 XT

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -913,10 +913,6 @@ if(LLGL_VK_ENABLE_SPIRV_REFLECT)
     message(STATUS "Including Submodule: SPIRV-Headers")
 endif()
 
-if(LLGL_VK_ENABLE_GPU_ASSISTED_VALIDATION)
-    message(STATUS "Enable Vulkan GPU assisted validation")
-endif()
-
 if(NOT "${SUMMARY_FLAGS}" STREQUAL "")
     message(STATUS "Options: ${SUMMARY_FLAGS}")
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -913,6 +913,10 @@ if(LLGL_VK_ENABLE_SPIRV_REFLECT)
     message(STATUS "Including Submodule: SPIRV-Headers")
 endif()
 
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -913,6 +913,10 @@ if(LLGL_VK_ENABLE_SPIRV_REFLECT)
     message(STATUS "Including Submodule: SPIRV-Headers")
 endif()
 
+if(LLGL_VK_ENABLE_GPU_ASSISTED_VALIDATION)
```
LukasBanana, line 916: This is not really needed here. I only added this extra status information for the SPIRV-Reflect option because it requires one of the few external dependencies.

LukasBanana (changes requested):
Thank you for your contribution. This looks good but I think we can skip the status information for this small option.

CMakeLists.txt
dolphineye74: Done

PR #219, opened 2025-07-24
 Fix Examples `MultiContext` for  Linux X11 OpenGL  

niu2x:
# Fix Examples `MultiContext` for  Linux X11 OpenGL  

# ​​Key Modifications Summary​​

- Independent GL Context per Window​
- Correct GL Context Sharing​
- call `MakeCurrent` Before Present
- `LinuxWindow::IsShown`  Implementation​

### DIFF
--- a/sources/Renderer/OpenGL/Platform/Linux/LinuxGLContextX11.cpp
+++ b/sources/Renderer/OpenGL/Platform/Linux/LinuxGLContextX11.cpp
@@ -338,7 +338,7 @@ GLXContext LinuxGLContextX11::CreateGLXContextCoreProfile(GLXContext glcShared, 
     return nullptr;
 }
 
-bool LinuxGLContextX11::IsSharableForSurface(Surface *surface)
+bool LinuxGLContextX11::IsSharableForSurface(const Surface *surface) const
 {
     return surface_ == surface;
 }
--- a/sources/Renderer/OpenGL/Platform/Linux/LinuxGLContextX11.h
+++ b/sources/Renderer/OpenGL/Platform/Linux/LinuxGLContextX11.h
@@ -41,7 +41,7 @@ class LinuxGLContextX11 : public LinuxGLContext
 
         OpenGL::RenderSystemNativeType GetNativeType() const override;
 
-        bool IsSharableForSurface(Surface *surface) override;
+        bool IsSharableForSurface(const Surface *surface) const override;
 
     public:
 


sources/Renderer/OpenGL/Platform/GLContextManager.cpp
```diff
@@ -143,17 +146,31 @@ std::shared_ptr<GLContext> GLContextManager::FindOrMakeContextWithPixelFormat(co
     /* Try to find pixel format with an exact match first */
     for (const GLPixelFormatWithContext& formatWithContext : pixelFormats_)
     {
-        if (formatWithContext.pixelFormat == pixelFormat)
+        if (formatWithContext.pixelFormat == pixelFormat) {
```
LukasBanana, line 149: If we really need to create one GLX context per window, we should do this via the base class `LLGL::GLContext`, e.g. with an interface function like `SupportsMultiDrawables()` or `IsSharableAcrossSurfaces()`.

That being said, I still think only one GLX context is necessary for multiple drawables. Here is a reference I found with a quick search: https://community.khronos.org/t/multiple-windows-single-opengl-context/77399
This also aligns with what ChatGPT suggested.

I also refreshed my memory about the MultiContext example, which calls `Present()` on both windows subsequently, so I think your change to make the context current in the `Present()` function might also be all that's needed. We should definitely explore this first since switching between multiple GL contexts is an expensive operation.

LukasBanana (changes requested):
Thank you for your first contribution, I'm glad to see some improvements in the Linux backend of LLGL.

I left some comments as I think we can simplify this a bit and keep to have only one GLX context for multiple windows. Please also update your formatting to _Allman style indentation_. I see if I can add a Clang format file soon to make this easier to follow.

sources/Renderer/OpenGL/Platform/Linux/LinuxGLContextX11.cpp
```diff
@@ -322,7 +324,7 @@ GLXContext LinuxGLContextX11::CreateGLXContextCoreProfile(GLXContext glcShared,
                 None
             };
 
-            GLXContext glc = glXCreateContextAttribsARB(display_, fbcList[0], nullptr, True, contextAttribs);
+            GLXContext glc = glXCreateContextAttribsARB(display_, fbcList[0], glcShared, True, contextAttribs);
```
LukasBanana, line 327: Good catch!

PR #222, opened 2025-08-20
Add a CMake option to disable LLGL windowing

st0rmbtw:


st0rmbtw:
> Is this causing compilation trouble or are there any binary size concerns?

No, it's not.

I just thought that if you use external windowing library then why would you include the code you don't use into the binary (specifically LLGL windowing code).

For example, there is a game engine called `bevy` that written Rust and it has fully modular structure. The user can add or remove unnecessary features like windowing, rendering system, etc. So, I thought it would be nice to have something like that here too. But, it seems that people don't do that kind of thing in C++.

Feel free to close this PR btw.

PR #226, opened 2025-11-30
Persistent map

swq0553:
**OpenGL Persistent Mapping**
**Persistent mapped buffers** in OpenGL, introduced with the ARB_buffer_storage extension in OpenGL 4.4, allow a buffer to remain mapped indefinitely. This technique is particularly useful for scenarios requiring frequent updates of data from the CPU to the GPU, such as particle systems or video frame updates. It reduces driver overhead and synchronization issues, making it a key component of the "Approaching Zero Driver Overhead" (AZDO) techniques.

**Persistent mapping** enables you to map a buffer once and retain the pointer for continuous updates without unmapping. This eliminates the need for repeated mapping and unmapping operations, reducing the number of OpenGL calls and synchronization overhead.

### DIFF
--- a/sources/Renderer/OpenGL/Buffer/GLBuffer.cpp
+++ b/sources/Renderer/OpenGL/Buffer/GLBuffer.cpp
@@ -61,20 +61,16 @@ GLBuffer::GLBuffer(long bindFlags, const char* debugName) :
 
 GLBuffer::~GLBuffer()
 {
-    if (mapped_)
+    #if LLGL_GLEXT_DIRECT_STATE_ACCESS
+    if (HasExtension(GLExt::ARB_direct_state_access))
     {
-        #if LLGL_GLEXT_DIRECT_STATE_ACCESS
-        if (HasExtension(GLExt::ARB_direct_state_access))
+        if (mapped_)
         {
             glUnmapNamedBuffer(GetID());
+            mapped_ = nullptr;
         }
-        else
-        #endif // /LLGL_GLEXT_DIRECT_STATE_ACCESS
-        {
-            GLStateManager::Get().BindGLBuffer(*this);
-            GLProfile::UnmapBuffer(GetGLTarget());
-        }  
     }
+    #endif // /LLGL_GLEXT_DIRECT_STATE_ACCESS
 
     glDeleteBuffers(1, &id_);
     GLStateManager::Get().NotifyBufferRelease(*this);
@@ -129,8 +125,6 @@ BufferDescriptor GLBuffer::GetDesc() const
             bufferDesc.cpuAccessFlags |= CPUAccessFlags::Read;
         if ((storageFlags & GL_MAP_WRITE_BIT) != 0)
             bufferDesc.cpuAccessFlags |= CPUAccessFlags::Write;
-        if ((storageFlags & GL_MAP_PERSISTENT_BIT) != 0)
-            bufferDesc.cpuAccessFlags |= CPUAccessFlags::Persistent;
     }
     else
     #endif // /GL_ARB_buffer_storage
@@ -151,13 +145,14 @@ void GLBuffer::BufferStorage(GLsizeiptr size, const void* data, GLbitfield flags
     #if LLGL_GLEXT_DIRECT_STATE_ACCESS
     if (HasExtension(GLExt::ARB_direct_state_access))
     {
-		storageFlags_ = flags;
+        // Immutable Buffer
+        storageFlags_ = flags;
         /* Allocate buffer with immutable storage (4.5+) */
         glNamedBufferStorage(GetID(), size, data, flags);
     }
     else
     #endif // /LLGL_GLEXT_DIRECT_STATE_ACCESS
-    #ifdef GL_ARB_buffer_storage
+    #if GL_ARB_buffer_storage
     if (HasExtension(GLExt::ARB_buffer_storage))
     {
         storageFlags_ = flags;
@@ -213,7 +208,7 @@ void GLBuffer::ClearBufferData(std::uint32_t data)
     }
     else
     #endif // /LLGL_GLEXT_DIRECT_STATE_ACCESS
-    #ifdef GL_ARB_clear_buffer_object
+    #if GL_ARB_clear_buffer_object
     if (HasExtension(GLExt::ARB_clear_buffer_object))
     {
         GLStateManager::Get().BindGLBuffer(*this);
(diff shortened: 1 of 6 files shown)


sources/Renderer/OpenGL/Buffer/GLBuffer.cpp
```diff
@@ -291,54 +306,106 @@ void GLBuffer::CopyBufferSubData(const GLBuffer& readBuffer, GLintptr readOffset
...
 void GLBuffer::UnmapBuffer()
 {
-    #if LLGL_GLEXT_DIRECT_STATE_ACCESS
-    if (HasExtension(GLExt::ARB_direct_state_access))
-    {
-        glUnmapNamedBuffer(GetID());
-    }
-    else
-    #endif // /LLGL_GLEXT_DIRECT_STATE_ACCESS
+    if (mapped_ == nullptr)
```
LukasBanana: I'm not a fan of auto-unmapping. Any call to `MapBuffer()` should be followed by an explicit call to `UnmapBuffer()`. This doesn't need to be an assertion that crashes the app, but I would just skip this here and report an error in the debug layer (i.e. in the destructor of `DbgBuffer`).

sources/Renderer/OpenGL/Buffer/GLBuffer.cpp
```diff
@@ -291,54 +306,106 @@ void GLBuffer::CopyBufferSubData(const GLBuffer& readBuffer, GLintptr readOffset
...
+                mapped_ = glMapNamedBufferRange(GetID(), 0, bufferSize_, flags);
+            }
+            else
+            #endif // /LLGL_GLEXT_DIRECT_STATE_ACCESS
+            {
+                GLStateManager::Get().BindGLBuffer(*this);
+                mapped_ = glMapBufferRange(GetGLTarget(), 0, bufferSize_, flags);
+            }
+        }
+        return (char*)mapped_ + offset;
```
LukasBanana: Please don't use C-style casts, they can unintentionally cast incompatible types, cast constness away, and silence compiler warnings. This can be a `static_cast` since we're casting from `void*` to something else.

sources/Renderer/OpenGL/Buffer/GLBuffer.cpp
```diff
@@ -291,54 +306,106 @@ void GLBuffer::CopyBufferSubData(const GLBuffer& readBuffer, GLintptr readOffset
...
+        {
+            return glMapNamedBufferRange(GetID(), offset, length, access);
+        }
+        else
+        #endif // /LLGL_GLEXT_DIRECT_STATE_ACCESS
+        #ifdef GL_ARB_map_buffer_range
+        if (HasExtension(GLExt::ARB_map_buffer_range))
+        {
+            GLStateManager::Get().BindGLBuffer(*this);
+            return glMapBufferRange(GetGLTarget(), offset, length, access);
```
LukasBanana: If we're calling this function at two places, here and above, we should move the direct call to `glMapBufferRange()` and the extension check into `GLProfile::MapBufferRange()` (in *Profile/GLCore/GLCoreProfile.cpp*). Its current implementation didn't support `GL_ARB_map_buffer_range`, but we should change that. The `GLProfile` namespace is meant to abstract the different GL profiles, because GLES only supports ranged-mapped buffers while desktop GL only supports it via extension.

sources/Renderer/OpenGL/Buffer/GLBuffer.cpp
LukasBanana: Does this pointer have to be reset in this function?

Also, does it make sense to effectively ignore this function if it is wrongfully invoked, i.e. shouldn't this be an error if someone calls `UnmapBuffer()` while it isn't mapped? Since this function is not directly called by the client, but via  `GLRenderSystem`, this should probably be an assertion, e.g. something like this:
```cpp
LLGL_ASSERT(mapped_ == nullptr, "Cannot unmap buffer that hasn't been mapped");
```

sources/Renderer/OpenGL/Buffer/GLBuffer.cpp
```diff
@@ -60,12 +55,27 @@ GLBuffer::GLBuffer(const BufferDescriptor& bufferDesc) :
...
 
-    if (bufferDesc.debugName != nullptr)
-        SetDebugName(bufferDesc.debugName);
+    if (debugName != nullptr)
+        SetDebugName(debugName);
 }
 
 GLBuffer::~GLBuffer()
 {
+    if (mapped_)
```
LukasBanana: I'm not a fan of auto-unmapping. Any call to `MapBuffer()` should be followed by an explicit call to `UnmapBuffer()`. This doesn't need to be an assertion here that crashes the app, but I would just skip this here and report an error in the debug layer (i.e. in the destructor of `DbgBuffer`).

tests/Testbed/UnitTests/TestBufferPersistentMap.cpp
```diff
@@ -0,0 +1,151 @@
...
+        }
+    }
+    else
+    {
+        Log::Errorf(LLGL::Log::ColorFlags::StdError, "Failed to map buffer 1 into CPU memory space for reading\n");
+        return TestResult::FailedErrors;
+    }
+
+    // Map buf2 into CPU memory space and write to
+    if (void* buf2Data = renderer->MapBuffer(*buf2, CPUAccess::WriteOnly))
```
LukasBanana: This test fails in D3D11/12:
```
Mismatch between data of buffer 2 feedback data [0x00000000, 0x00000000, 0x00000000, 0x00000000] and initial data [0x00000012, 0x00000034, 0x00000056, 0x00000078]
================================================================================
   Test BufferPersistentMap: [ FAILED - MISMATCH ]
```

In GL, this test also reports issues in the debug layer:
```
error (invalid state): in 'ReadBuffer': reading uninitialized buffer
error (invalid state): in 'MapBuffer': cannot map buffer that has already been mapped to CPU memory space
warning (improper argument): in 'CreateBuffer': unknown CPU access flags specified for buffer
================================================================================
   Test BufferPersistentMap: [ Ok ]
```
This might need an update in `DbgRenderSystem` and/or `DbgBuffer` to account for the new feature.

sources/Renderer/OpenGL/Buffer/GLBuffer.cpp
swq0553: mapped is for persistent map. default is nullptr,when not nullptr indicate that it is in peristent mapping.
for persistent map,we just map once and unmap it when destory.
so in GLBuffer::unmap we dont do really unmap.

swq0553:
persistent map not work well for some case,
eg:In GLDeferredCommandBuffer the buffer is same with the last change.
but one can change many time and do some work use this buffer.

PR #235, opened 2026-05-14
Add open xr support

joevdh:
Add support for OpenXR, which has to integrate with each graphics API as part of their initialization and swap chain creation.  Support for Vulkan, D3D11 and D3D12 only, tested on Windows and Android XR headsets (Vulkan only)

- Adds XRSystem as a peer of RenderSystem
- XRSystem manages the initialization of OpenXR, and initializes the graphics devices, which are then passed to existing render initialization code.
- XR swapchains are supported through a new interface to LLGL::Texture that allows them to be set up outside the existing rendering code and passed in
- Added HelloOpenXR example app


LukasBanana:
Thank you for your contribution!

This is quite a large PR but I'll try to get through the review over the weekend.

Can I assume your second PR builds on-top of this one? If so, it probably makes sense to get this PR integrated before I start reviewing the second one, which also seems quite large.

Laura

joevdh:
Yes, the second one builds on the first, but also has some issues that came up while I was trying to use it locally.  I put the other PR as a draft while I sort it out.

examples/Cpp/HelloOpenXR/Example.450core.frag.spv
LukasBanana, line 1: Do these need to be compiled in a specific way? You can add new SPIR-V shaders to *examples/Cpp/CompileAllGlslExamplesToSpirv.bat*.

examples/Cpp/HelloOpenXR/Example.cpp
```diff
@@ -0,0 +1,700 @@
...
+#   define LLGL_HELLO_LOG(...) std::printf(__VA_ARGS__)
+#   define LLGL_HELLO_ERR(...) std::fprintf(stderr, __VA_ARGS__)
+#endif
+
+
+// -----------------------------------------------------------------------------
+// Math helpers (column-major 4x4, OpenXR right-handed +Y up / -Z forward)
+// -----------------------------------------------------------------------------
+
+struct Mat4
```
LukasBanana: For the examples, I use my own library [GaussianLib](https://github.com/LukasBanana/GaussianLib) project for matrices and vectors. While this library is optional for LLLG itself, it is required for the examples, so we should make use of it instead of introducing new matrix types and re-implementing its arithmetic operators. You can include it via `<Gauss/Matrix.h>` and use `Gs::Matrix4f` for a 4x4 float matrix.

examples/Cpp/HelloOpenXR/Example.cpp
```diff
@@ -0,0 +1,700 @@
...
+                s += a.m[k * 4 + row] * b.m[col * 4 + k];
+            r.m[col * 4 + row] = s;
+        }
+    }
+    return r;
+}
+
+// View matrix that takes world-space points to camera space, given the XR pose
+// (camera-to-world transform = translate(position) * rotate(orientation)).
+static Mat4 ViewFromPose(const LLGL::XRViewPose& pose)
```
LukasBanana: With `Gs::Matrix4f`, you can use the same subscript operator as with C-style arrays.

examples/Cpp/HelloOpenXR/Example.cpp
```diff
@@ -0,0 +1,700 @@
...
+    return p;
+}
+
+
+// -----------------------------------------------------------------------------
+// Cube geometry (axis-coloured: +X red, -X cyan, +Y green, -Y magenta,
+//                                +Z blue, -Z yellow)
+// -----------------------------------------------------------------------------
+
+struct CubeVertex
```
LukasBanana: The examples usually use `TexturedVertex` and generate a cube with `GenerateTexturedCubeVertices()` from *ExampleBase/GeometryUtils.h*. I would keep using a textured cube with basic lighting (you can adopt those shaders from *examples/Cpp/Texturing/)*. If you want to keep just rendering a solid colored cube, I suggest adding a new struct to *GeometryUtils.h* named `ColoredVertex` and either convert the textured cube vertices or add a new function in those utility files, so other examples can use them, too. This keeps the boilerplate code for geometry generation out of the examples and makes them a bit more concise.

examples/Cpp/HelloOpenXR/Example.cpp
```diff
@@ -0,0 +1,700 @@
...
+    { { -kS, +kS, +kS }, { 0.0f, 0.0f, 1.0f } },
+    { { -kS, -kS, +kS }, { 0.0f, 0.0f, 1.0f } },
+    // -Z face (yellow)
+    { { -kS, -kS, -kS }, { 1.0f, 1.0f, 0.0f } },
+    { { -kS, +kS, -kS }, { 1.0f, 1.0f, 0.0f } },
+    { { +kS, +kS, -kS }, { 1.0f, 1.0f, 0.0f } },
+    { { +kS, -kS, -kS }, { 1.0f, 1.0f, 0.0f } },
+};
+
+static constexpr std::uint16_t kCubeIndices[36] =
```
LukasBanana: The same index sequence is already provided by `GenerateTexturedCubeTriangleIndices()` in *GeometryUtils.cpp*. Even if you use a different vertex struct, you can use the existing indices generator.

examples/Cpp/HelloOpenXR/Example.cpp
```diff
@@ -0,0 +1,700 @@
...
+// dropped 30 cm below eye height so it's natural to look down on it slightly.
+static const float kCubeWorldOffset[3] = { 0.0f, -0.30f, -1.5f };
+
+
+// -----------------------------------------------------------------------------
+// File I/O helper
+// -----------------------------------------------------------------------------
+
+#if defined LLGL_OS_ANDROID
+static AAssetManager* g_androidAssetManager = nullptr;
```
LukasBanana: The Android asset manager reference is also already handled in ExampleBase in *FileUtils.h*/*.cpp* and *examples/Shared/Platform/Android/*.

examples/Cpp/CMakeLists.txt
```diff
@@ -137,6 +141,11 @@ if(LLGL_BUILD_EXAMPLES)
             add_llgl_example_project(Example_MultiContext   CXX "${FilesExample_MultiContext}"      "${EXAMPLE_PROJECT_LIBS}")
             add_llgl_example_project(Example_MultiRenderer  CXX "${FilesExample_MultiRenderer}"     "${EXAMPLE_PROJECT_LIBS}")
         endif()
+
+        # OpenXR example - drives an XR runtime, no display window or GaussianLib needed.
+        if(LLGL_BUILD_XR_OPENXR)
+            add_llgl_example_project(Example_HelloOpenXR    CXX "${FilesExample_HelloOpenXR}"       "LLGL_XR_OpenXR;${EXAMPLE_PROJECT_LIBS}")
```
LukasBanana, line 147: With `EXAMPLE_PROJECT_LIBS`, you already link against *ExampleBase* btw.

include/LLGL/XR/XRSession.h
```diff
@@ -0,0 +1,190 @@
...
+    std::int64_t                predictedDisplayTimeNs  = 0;
+
+    //! True if the application should render this frame. If false, BeginFrame/EndFrame must still be paired but no rendering work is needed.
+    bool                        shouldRender            = false;
+
+    //! Per-view pose and projection parameters sampled at frame begin. Size matches XRSystem::GetViewConfigurations().
+    DynamicVector<XRViewPose>   views;
+
+    /**
+    \brief Near clipping plane (in metres) the application used to render this frame.
```
LukasBanana: Minor nitpick: Please use `en-US` locale, "meter" instead of "metre".

include/LLGL/XR/XRSystemFlags.h
```diff
@@ -0,0 +1,274 @@
...
+/* ----- Structures ----- */
+
+/**
+\brief XR system descriptor structure.
+\see XRSystem::Load
+*/
+struct XRSystemDescriptor
+{
+    //! Application name reported to the OpenXR runtime. Defaults to "LLGL Application".
+    UTF8String              applicationName     = "LLGL Application";
```
LukasBanana: The first four fields can be replaced by `LLGL::ApplicationDescriptor`. If you need them to use `UTF8String` instead of C-style strings, we can discuss whether it makes sense to change that in that existing struct, but in that case I would rather make them `LLGL::StringLiteral`. For reference, `LLGL::RendererConfigurationVulkan` uses it so far.

include/LLGL/XRSystemFlags.h
```diff
@@ -0,0 +1,274 @@
...
+
+    #endif // /LLGL_OS_ANDROID
+};
+
+/**
+\brief XR render system descriptor structure.
+\remarks Used to obtain an LLGL::RenderSystem whose underlying device satisfies the XR runtime's requirements.
+\see XRSystem::CreateRenderSystem
+*/
+struct XRRenderSystemDescriptor
```
LukasBanana: Can we please replace this struct and use the existing `RenderSystemDescriptor`? This looks redundant. With this struct, I don't see how the client can pass in an instance of `LLGL::RenderingDebugger` as validation layer. The client should also still be able to pass in their own native handle. If that was an issue, you could simply add an assertion or return a failure (via the `Report` handle) if the input native handle is non-null and then override it in your `*OpenXRGraphicsBinding::CreateRenderSystem()` implementations. I'd like to keep the number of newly introduced interfaces and structs small so we can keep the overall LLGL interface concise.

include/LLGL/XR/XRSystemFlags.h
```diff
@@ -0,0 +1,274 @@
...
+};
+
+/**
+\brief Per-view configuration entry returned by XRSystem::GetViewConfigurations.
+\see XRSystem::GetViewConfigurations
+*/
+struct XRViewConfigurationView
+{
+    //! Recommended image width (in pixels) for this view's swap-chain.
+    std::uint32_t   recommendedImageWidth   = 0;
```
LukasBanana: Would you mind if we combine the `Width`/`Height` fields by using `LLGL::Extent2D` and call these fields `*ImageExtent` or `*ImageSize`? Same as with `resolution`, it makes it clear that these fields should always be set together.

sources/Renderer/Direct3D11/OpenXR/D3D11OpenXRGraphicsBinding.cpp
```diff
@@ -0,0 +1,404 @@
...
+namespace OpenXR
+{
+
+
+static const char* g_requiredExtensions[] =
+{
+    XR_KHR_D3D11_ENABLE_EXTENSION_NAME,
+};
+
+static Format DxgiColorFormatToLLGL(DXGI_FORMAT format)
```
LukasBanana: `DxgiColorFormatToLLGL()` and `DxgiDepthFormatToLLGL()` can be replace with `DXTypes::Unmap()` and `LLGLToDxgiFormat()` can be replaced by `DXTypes::Map()`. If you need these to also validate a format is only a color and depth-stencil format, you can use `LLGL::IsColorFormat()` and `LLGL::IsDepthOrStencilFormat()` respectively.

sources/Renderer/Direct3D11/OpenXR/D3D11OpenXRGraphicsBinding.cpp
```diff
@@ -0,0 +1,404 @@
...
+    return renderSystem.GetNativeHandle(&outHandle, sizeof(outHandle));
+}
+
+
+D3D11OpenXRGraphicsBinding::D3D11OpenXRGraphicsBinding() = default;
+D3D11OpenXRGraphicsBinding::~D3D11OpenXRGraphicsBinding() = default;
+
+const char* D3D11OpenXRGraphicsBinding::GetRendererModuleName() const
+{
+    return "Direct3D11";
```
LukasBanana, line 47: I'm not a fan of having the module name hard coded twice. The declaration via `LLGL_IMPLEMENT_RENDERER_MODULE()` in **ModuleInterface.cpp* should be the single-source of those names. I don't have a good solution for this at the moment, though. I think the way `XRSystem` adds bindings from backend specific implementation is a similar circular dependency as LLGL's static linking where the core library needs to link against its backends in order to instantiate them. Not optimal, but maybe something we should fix separately.
For now, I think we can keep it as-is, since the whole `GraphicsBinding` is not in the public interface, so we can change it any time.

sources/Renderer/Direct3D11/OpenXR/D3D11OpenXRGraphicsBinding.cpp
```diff
@@ -0,0 +1,404 @@
...
+}
+
+ArrayView<const char*> D3D11OpenXRGraphicsBinding::GetRequiredXrExtensions() const
+{
+    return ArrayView<const char*>{ g_requiredExtensions, sizeof(g_requiredExtensions) / sizeof(g_requiredExtensions[0]) };
```
LukasBanana: You can either use the `LLGL_ARRAY_LENGTH()` macro from *MacroUtils.h* for the second expression or just remove the second parameter altogether, since `ArrayView` as a constructor for fixed-size arrays.

sources/Renderer/Direct3D11/OpenXR/D3D11OpenXRGraphicsBinding.cpp
```diff
@@ -0,0 +1,404 @@
...
+        return nullptr;
+    }
+
+    // Hand the prepared device to LLGL's D3D11 backend via the existing nativeHandle path.
+    Direct3D11::RenderSystemNativeHandle nativeHandle{};
+    nativeHandle.device         = d3dDevice.Get();
+    nativeHandle.deviceContext  = d3dContext.Get();
+
+    RenderSystemDescriptor rsDesc;
+    rsDesc.moduleName           = "Direct3D11";
```
LukasBanana: Once we replace `XRRenderSystemDescriptor`, `moduleName` will already contain the correct name, since it was used to select the `GraphicsBinding` instance.

sources/Renderer/Direct3D11/OpenXR/D3D11OpenXRGraphicsBinding.cpp
```diff
@@ -0,0 +1,404 @@
...
+        entry.texture = texture;
+        outImages.push_back(entry);
+    }
+    if (native.device) native.device->Release();
+    return true;
+}
+
+void D3D11OpenXRGraphicsBinding::ReleaseSwapchainImage(RenderSystem& /*renderSystem*/, Texture& texture)
+{
+    delete static_cast<D3D11Texture*>(&texture);
```
LukasBanana: Who is the owner of those textures? We need an interface that owns this memory which should use `std::unique_ptr` and `LLGL::MakeUnique<>()` so we can avoid manual memory deallocation. We should only use `delete` for Pimpl idioms like in `Blob::~Blob()` where it's always guaranteed to be freed. Who guarantees this function is called? If you forget to call `RenderSystem::Release(Texture&)` for instane, the render system will still clear up all textures because they are all stored as unique pointers.

sources/Renderer/Direct3D11/OpenXR/D3D11OpenXRGraphicsBinding.cpp
```diff
@@ -0,0 +1,404 @@
...
+    const D3D_FEATURE_LEVEL featureLevels[] =
+    {
+        D3D_FEATURE_LEVEL_11_1,
+        D3D_FEATURE_LEVEL_11_0,
+    };
+
+    ComPtr<ID3D11Device> d3dDevice;
+    ComPtr<ID3D11DeviceContext> d3dContext;
+    D3D_FEATURE_LEVEL chosenLevel = D3D_FEATURE_LEVEL_11_0;
+    hr = D3D11CreateDevice(
```
LukasBanana: What is this D3D11 device creation doing that is missing in `D3D11RenderSystem::CreateDevice()`? Is this primarily to satisfy the `GraphicsBinding` interface because the Vulkan backends needs to do a lot more?

My concern here is that this can easily diverge from the default device creation and in the best case, we have to maintain two implementations that handle this D3D device setup.

sources/Renderer/Direct3D11/Texture/D3D11Texture.h
```diff
@@ -37,6 +37,23 @@ class D3D11Texture final : public Texture
 
         D3D11Texture(ID3D11Device* device, const TextureDescriptor& desc);
 
+        // Adoption parameters for wrapping an externally-owned ID3D11Texture2D
+        // (e.g. an XR-runtime-owned swap-chain image).
+        struct AdoptionParams
```
LukasBanana: I think this struct can be boiled down to a single input parameter `ID3D11Texture2D*` for the secondary constructor and call it something like `externalTex2D`.
Most of these fields don't seem to be used in the new constructor, you use `ID3D11Texture2D::GetDesc()` and use those "actual" parameters, which is the right way to do it IMO. `llglFormat` can then be derived via `DXTypes::Unmap()`, `type` and `bindFlags` must be derived from the information we have from `D3D11_TEXTURE2D_DESC`. We can do that in a function like this:
```cpp
static void GetD3DTexture2DInfo(const D3D11_TEXTURE2D_DESC& inDesc, TextureType& outType, long& outBindFlags, Format& outFormat)
{
    /* Determine texture type */
    if ((inDesc.MiscFlags & D3D11_RESOURCE_MISC_TEXTURECUBE) != 0 && inDesc.ArraySize % 6 == 0)
    {
        if (inDesc.ArraySize > 6)
            outType = TextureType::TextureCubeArray;
        else
            outType = TextureType::TextureCube;
    }
    else if (inDesc.SampleDesc.Count > 1)
    {
        if (inDesc.ArraySize > 1)
            outType = TextureType::Texture2DMSArray;
        else
            outType = TextureType::Texture2DMS;
    }
    else
    {
        if (inDesc.ArraySize > 1)
            outType = TextureType::Texture2DArray;
        else
            outType = TextureType::Texture2D;
    }

    /* Determine binding flags */
    // See DXGetBufferBindFlags() and DXGetTextureBindFlags() how to unmap from D3D to LLGL

    /* Map DXGI to LLGL format */
    outFormat = DXTypes::Unmap(inDesc.Format);
}
```

sources/Renderer/Direct3D12/OpenXR/D3D12OpenXRGraphicsBinding.cpp
```diff
@@ -0,0 +1,388 @@
...
+namespace OpenXR
+{
+
+
+static const char* g_requiredExtensions[] =
+{
+    XR_KHR_D3D12_ENABLE_EXTENSION_NAME,
+};
+
+static Format DxgiColorFormatToLLGL(DXGI_FORMAT format)
```
LukasBanana: This is the duplication we can avoid if we use `DXTypes`, which is defined in the *DXCommon* static library and linked into Direct3D11 and Direct3D12 modules.

sources/Renderer/Direct3D12/Texture/D3D12Texture.h
```diff
@@ -37,6 +37,19 @@ class D3D12Texture final : public Texture
 
         D3D12Texture(ID3D12Device* device, const TextureDescriptor& desc);
 
+        // Adoption parameters for wrapping an externally-owned ID3D12Resource
+        // (e.g. an XR-runtime-owned swap-chain image).
+        struct AdoptionParams
```
LukasBanana: I think we can do the same reduction here as in D3D11 and only use a single parameter `ID3D12Resource* externalTex2D` and derive the remaining necessary parameters. Maybe `initialState` is the exception, but if this is only used internally, we already know what the initial state will be.

sources/Renderer/Vulkan/OpenXR/VKOpenXRGraphicsBinding.cpp
```diff
@@ -0,0 +1,480 @@
...
+        case Format::RGBA16Float:           return VK_FORMAT_R16G16B16A16_SFLOAT;
+        case Format::D16UNorm:              return VK_FORMAT_D16_UNORM;
+        case Format::D24UNormS8UInt:        return VK_FORMAT_D24_UNORM_S8_UINT;
+        case Format::D32Float:              return VK_FORMAT_D32_SFLOAT;
+        case Format::D32FloatS8X24UInt:     return VK_FORMAT_D32_SFLOAT_S8_UINT;
+        default:                            return VK_FORMAT_UNDEFINED;
+    }
+}
+
+static VkSampleCountFlagBits SampleCountToVkBits(std::uint32_t sampleCount)
```
LukasBanana: I'm not sure if you copy/pasted this from `VKTypes::ToVkSampleCountBits()` due to linker issues, but this also looks like unnecessary duplicated boilerplate code.

sources/Renderer/Vulkan/OpenXR/VKOpenXRGraphicsBinding.cpp
```diff
@@ -0,0 +1,480 @@
...
+namespace OpenXR
+{
+
+
+static const char* g_requiredExtensions[] =
+{
+    XR_KHR_VULKAN_ENABLE2_EXTENSION_NAME,
+};
+
+static Format VkColorFormatToLLGL(VkFormat format)
```
LukasBanana: Same as with D3D, this should use functions from `VKTypes`.

sources/Renderer/Vulkan/OpenXR/VKOpenXRGraphicsBinding.cpp
```diff
@@ -0,0 +1,480 @@
...
+}
+
+ArrayView<const char*> VKOpenXRGraphicsBinding::GetRequiredXrExtensions() const
+{
+    return ArrayView<const char*>{ g_requiredExtensions, sizeof(g_requiredExtensions) / sizeof(g_requiredExtensions[0]) };
+}
+
+bool VKOpenXRGraphicsBinding::LoadFunctions(XrInstance instance, Functions& fns, Report* report)
+{
+    auto get = [&](const char* name, PFN_xrVoidFunction* outPfn) -> bool
```
LukasBanana, line 100: Can we please rename this to `LoadXRProc` to match the naming in *VKExtensionLoader.cpp*?

sources/Renderer/Vulkan/Texture/VKDeviceImage.h
```diff
@@ -58,6 +58,10 @@ class VKDeviceImage
             VkImageUsageFlags       usageFlags
         );
 
+        // Adopts an externally-owned VkImage. The image is held as a weak reference;
+        // this VKDeviceImage will not destroy it.
+        void AdoptVkImage(VkImage externalImage, VkImageLayout initialLayout);
```
LukasBanana: Let's call this `TakeVkImageOwnership` or `MoveVkImage` to highlight that we move ownership of the `VkImage`, since `VKDeviceImage` will delete this in its destructor. With "adopt" I am thinking more of a copy as in adopting an idea by copying it, but this actually transfers ownership and responsibility of who's meant to free the memory.

sources/Renderer/Vulkan/Texture/VKTexture.h
```diff
@@ -46,6 +46,25 @@ class VKTexture final : public Texture
             const TextureDescriptor&    desc
         );
 
+        // Adoption constructor. Wraps an externally-owned VkImage (e.g. from an XR runtime
+        // or another swap-chain implementation). The VkImage is not destroyed by this object.
+        // The caller is responsible for ensuring the image outlives this texture.
+        struct AdoptionParams
```
LukasBanana: With Vulkan, we have to provide a lot more parameters than just the handle indeed, but let's move this outside of `VKTexture` and into `LLGL` namespace and call it `VKTextureDescriptor` or `VKExternalImageInfo`. With internal classes, I usually keep its structs in the `LLGL` namespace, for example `D3D11StaticSampler` and `D3D11Sampler`.

sources/XR/OpenXR/OpenXRError.h
```diff
@@ -0,0 +1,109 @@
...
+    report->Errorf("%s failed: %s\n", contextLabel, ResultToString(instance, result));
+}
+
+// Reports an OpenXR failure into the given Report object using a printf-style label.
+inline void ReportXrErrorf(Report* report, XrInstance instance, XrResult result, const char* format, ...)
+{
+    if (report == nullptr)
+        return;
+
+    char prefix[256];
```
LukasBanana: Is there an OpenXR macro or where does this number come from? If we can guarantee this is the maximum required buffer size, I'm okay with this. Otherwise, we can use `LLGL_STRING_PRINTF()`, which uses two varg lists to first determine the required size for an `std::string` and then fills in the data.

sources/XR/OpenXR/OpenXRGraphicsBinding.h
```diff
@@ -0,0 +1,169 @@
...
+namespace OpenXR
+{
+
+
+/**
+\brief Per-image entry returned by GraphicsBinding::EnumerateSwapchainImages.
+\remarks The texture is owned by the binding (created via the underlying LLGL::RenderSystem) and is released
+when the OpenXRSwapChain is destroyed.
+*/
+struct SwapchainImage
```
LukasBanana: This name should be consistent with `XRSwapChain`. Do you mind naming this `XRSwapChainImage`?

sources/XR/OpenXR/OpenXRSession.cpp
```diff
@@ -0,0 +1,452 @@
...
+    ));
+
+    OpenXRSwapChain* raw = swapChain.get();
+    ownedSwapChains_.push_back(std::move(swapChain));
+    return raw;
+}
+
+void OpenXRSession::Release(XRSwapChain& swapChain)
+{
+    for (auto it = ownedSwapChains_.begin(); it != ownedSwapChains_.end(); ++it)
```
LukasBanana: If you use `HWObjectContainer<XRSwapChain>` for `ownedSwapChains_`, similarly to all the renderer containers, you can simplify this function to just `ownedSwapChains_.erase(&swapChain)`. This erase also has complexity `O(1)`, because it stores its own index as payload in front of the entry and uses a swap-with-last-entry algorithm (a bit similar to `std::make_shared`, which stores the ref-counter as part of the primary memory allocation). This won't make a big performance differnece, this I presume we'll only ever have two swap-chains (one for each eye), but it makes this function simpler and unifies the containers we use for instance ownership.

BuildAndroid.sh
```diff
@@ -277,13 +308,24 @@ generate_app_project()
...
+    # XR projects mark themselves with an Android.openxr file. When present, swap in the XR
+    # manifest variant (immersive_hmd feature, OPENXR permissions, broker queries, intent
+    # category) and remember to bundle the loader .so below.
+    NEEDS_OPENXR_LOADER=0
+    if [ -f "$PROJECT_SOURCE_DIR/Android.openxr" ]; then
+        NEEDS_OPENXR_LOADER=1
+        cp "$PLATFORM_SOURCE_DIR/app/src/main/AndroidManifest.xr.xml" "$APP_ROOT/app/src/main/AndroidManifest.xml"
+    fi
+    # Always remove the XR manifest variant from the copied tree - it's a template, not a build artifact.
+    rm -f "$APP_ROOT/app/src/main/AndroidManifest.xr.xml"
```
LukasBanana: Copying the whole folder tree, overriding an existing file, and then deleting the unwanted one seems like a hassle. Do you think it's nicer to have a separate folder tree which only contains the XR version of the manifest and we just copy that to override the other XML file? I.e. we have *app/* as before and also *app-xr/src/main/AndroidManifest.xml*. This way we can remove this `rm` statement and do the override without having to rename the file, we just copy it over the default one in the destination folder.

BuildAndroid.sh
```diff
@@ -200,7 +217,9 @@ build_with_android_abi()
     fi
 
     if [ $PROJECT_ONLY -eq 0 ]; then
-        cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE ${OPTIONS[@]}
+        # Force Ninja: CMake's default on Windows is Visual Studio, which can't cross-compile
+        # to Android via the NDK toolchain file.
+        cmake -G Ninja -DCMAKE_BUILD_TYPE=$BUILD_TYPE ${OPTIONS[@]}
```
LukasBanana, line 222: Nice catch! I have to admit that my setup on Windows might be a bit of a niche when building for Android. I use MSYS2.

CMakeLists.txt
```diff
@@ -180,6 +180,13 @@ if(WIN32)
     option(LLGL_BUILD_RENDERER_DIRECT3D12 "Include Direct3D12 renderer project" ON)
 endif()
 
+# OpenXR support (requires at least one compatible renderer; currently only Vulkan).
+# Available on desktop (Windows/Linux/macOS) and Android (e.g. Quest, HTC Vive Focus).
+# iOS has no OpenXR runtime, UWP and WASM use different XR APIs.
+if(NOT LLGL_IOS_PLATFORM AND NOT LLGL_UWP_PLATFORM AND NOT EMSCRIPTEN)
+    option(LLGL_BUILD_XR_OPENXR "Include OpenXR support (LLGL::XRSystem). Requires LLGL_BUILD_RENDERER_VULKAN for the Vulkan binding." OFF)
```
LukasBanana: I guess the second statement is outdated since this also works with D3D11 and D3D12 now?

include/LLGL/InterfaceID.h
```diff
@@ -54,6 +54,9 @@ struct InterfaceID
         Shader,                 //!< Extends RenderSystemChild. \see Shader
         RenderTarget,           //!< Extends RenderSystemChild. \see RenderTarget
         SwapChain,              //!< Extends RenderTarget. \see SwapChain
+        XRSystem,               //!< Extends Interface. \see XRSystem
```
LukasBanana: Let's move this after `RenderSystemChild` and before `BufferArray` so that those are the last entries in this enum that are direct subclasses of `Interface`. Not mandatory for the `IsInstanceOf()` system to work, as long as the Ids are higher than the one for `Interface`, but it follows the same pattern of the other entries here.

LukasBanana (changes requested):
First of all, great work! This is obviously a big change proposal but I think I went through most of it and left several comments. Here are my main points I'd like to adjust:

1. I would like us to make use of ExampleBase for the HelloOpenXR example as I see quite a bit of boilerplate code we can avoid. I noticed you added a custom matrix type, but since the examples already require the third party library GaussianLib (not required for LLGL itself), we should make use of it, so we don't have to define basic matrix operations.
Besides this, I noticed some duplication from `DXTypes` and `VKTypes` namespaces. We should consolidate those format and type mapping functions to reduce code duplication.
2. I like your overall folder structure and I'm fine with introducing a new category under *sources/XR/*. Also the introduction of *OpenXR/* under the renderer backends makes sense to me. We can also keep the `LLGL::XR*` identifiers as I think we don't need to wrap them in a new namespace if we use the `XR` prefix. The only thing I'd like to change is to move the XR header files from *include/LLGL/* into *include/LLGL/XR/*. This way it's clear they are not required for the core interface of LLGL, analogous to the *Utils/* folder, and are separated since they are acting like an extension.
3. I'd appreciate it if we can replace the `XRRenderSystemDescriptor` struct. This seems redundant and I think we can utilize the existing `RenderSystemDescriptor` struct. I added some more info to this in the respective section.
Similarly to this, I also think we should avoid the `AdoptionParams` structs where we can. Most of those fields are not even used and we can derive all the required information from the D3D `GetDesc()` functions. This way we have a single source of truth via the externally allocated texture handle. Otherwise, you can accidentally pass in a DXGI 2D texture but provide parameters that match those of a cube texture.

I can't say much about the `XRSession` and `OpenXRSystem` implementations, so I'm just going to trust your judgement here. After some issue with the *jsoncpp* CMake script on my machine (it kept pointing me to the MSYS2 folder, which collides with MSVC), I was able to compile and run successfully, but I don't have a VR setup on my PC  at the moment; although I do have a PSVR2, which I know can somehow be connected to PC these days.

Thanks again for your first contribution, I also appreciate you took care of the build scripts :)

joevdh:
So for the example, ExampleBase does all the initialization of RenderSystem, the swap chain, etc in the constructor, but when using OpenXR that needs to be done differently.  So to use ExampleBase I'd have to either make a new constructor that doesn't do any initialization so the HelloOpenXR example can do the setup itself, or add an option to perform the XR compatible initialization in ExampleBase but this would mean pushing some XR specific code into the base that all examples use.   

How would you like to approach this?

LukasBanana:
I mean you don't have to use the `ExampleBase` class, just link against the library (I think you already do) and use the utility functions and matrix types I was referring to. The *MultiRenderer* example doesn't use the constructor either because it creates multiple render systems and can't use the default initialization, but still use the vertex generation functions (see [examples/Cpp/MultiRenderer/Example.cpp:436](https://github.com/LukasBanana/LLGL/blob/master/examples/Cpp/MultiRenderer/Example.cpp#L436))

joevdh:
Oh!  Ok, I misunderstood what you meant.  Sound good

joevdh:
Addressed all review feedback and squashed into a single commit ([459ac0c](https://github.com/LukasBanana/LLGL/commit/459ac0ca07ee66075e1b7b98fcce63d6e173c3e2)).  Have a look and see what you think.

include/LLGL/XR/XRSystemFlags.h
LukasBanana: Can you please revise this part. I still think we should consolidate these first four fields and use the existing `ApplicationDescriptor`. You can find it by including `<LLGL/RendererConfiguration.h>`.

include/LLGL/XR/XRSystemFlags.h
LukasBanana: Any opinion on this part? I'd prefer to make use of the existing `Extent2D` structs here.

sources/Renderer/Direct3D11/OpenXR/D3D11OpenXRGraphicsBinding.cpp
LukasBanana: We should really consolidate these mapping functions by using the existing ones in `DXTypes`. Same for D3D12 and Vulkan.

LukasBanana (changes requested):
The HelloOpenXR example looks good. Thanks for the update.

Could you please take a look at my feedback in the **GraphicsBinding.cpp* and **Texture.cpp* source files? I think we can consolidate a little bit more with respect to type mapping. Most of that should be handled by the `DXTypes` and `VKTypes` namespaces.

The `AdoptionParams` can be replaced as well (except for Vulkan). For D3D we can use `GetDesc()` and have the input handle (e.g. `ID3D11Resource`) as single source of truth for those parameters. I think that's important when LLGL has to rely on externally owned textures so that the parameters are guaranteed to be in sync.

sources/Renderer/Direct3D11/OpenXR/D3D11OpenXRGraphicsBinding.cpp
joevdh: Ah!  Totally missed this the first time around.  Will fix

include/LLGL/XR/XRSystemFlags.h
joevdh: no prob!

sources/Renderer/Direct3D11/OpenXR/D3D11OpenXRGraphicsBinding.cpp
LukasBanana: @joevdh do you have some insight for this question? Perhaps this went under your radar.

sources/Renderer/Direct3D11/Texture/D3D11Texture.h
LukasBanana: @joevdh  I still don't see any response to my points here. Can you please take a look at this?

sources/Renderer/Direct3D12/Texture/D3D12Texture.h
LukasBanana: @joevdh Here as well.

sources/Renderer/Vulkan/VKRenderSystem.cpp
```diff
@@ -984,7 +1031,7 @@ static VKAPI_ATTR VkBool32 VKAPI_CALL VKDebugCallback(
     if ((flags & VK_DEBUG_REPORT_ERROR_BIT_EXT) != 0)
     {
         VKRenderSystem* renderSystemVK = static_cast<VKRenderSystem*>(userData);
-        if (renderSystemVK->IsBreakOnErrorEnabled())
+        if (renderSystemVK->IsBreakOnErrorEnabled() && !IsKnownXrRuntimeFalsePositive(message))
```
LukasBanana: Should this only be filtered out when the OPENXR macro is defined?

LukasBanana (changes requested):
I skimmed over the review. I still see you haven't responded to some of my questions. Can you please go over them to see if you have some insight for those points? Let me know if you need help with some of those, like replacing the `AdoptionParams` struct.
Please note that I'll be out of town all of next week, so I might respond only sporadically.

sources/Renderer/Direct3D11/OpenXR/D3D11OpenXRGraphicsBinding.cpp
joevdh: Good call, that setup was awkward.  The idea was the lifetime of the textures matched the lifetime of the OpenXRSwapChain that used them, but each texture is specific to a different rendering backend, so the OpenXRSwapChain had to rely on the binding to make the textures for it.  But this function isn't needed, Textures have virtual destructors.  I've updated XRSwapChainImage to have a std::unique_ptr to make the ownership more explicit, and switched to using a std::vector to hold the array of images since SmallVector requires copyable types.  Does that work?

sources/Renderer/Direct3D11/Texture/D3D11Texture.h
joevdh: True, most of these parameters are unused.  But the only (protected) constructor for the Texture base class requires the TextureType and bindFlags, so we can't just pass just the ID3D11Texture2D*.  

So, I can cull all the unneeded values from the Adoption params except the ID3D11Texture2D, TextureType, and bindFlags and derive the rest in the constructor as you suggest, or we can modify the base class to not require the TextureType and bindFlags as construction parameters.  Which approach would you prefer?  I'm leaning towards the smaller AdoptionParams option, since the RAII-nature of the constructor for Texture seems like a good thing to keep and enforce.  Thoughts?

sources/Renderer/Vulkan/VKRenderSystem.cpp
joevdh: Yeah, good call

joevdh: I hate that this is even necessary

sources/XR/OpenXR/OpenXRError.h
joevdh: No, the variadic args in this case are coming from LLGL implementation code, not OpenXR.  I'll switch to LLGL_STRING_PRINTF

sources/Renderer/Direct3D11/Texture/D3D11Texture.h
LukasBanana: If you don't want to go down the route I suggested in the above code snippet with `GetD3DTexture2DInfo()` that's fine, then just trim down the `AdotionParams` struct and I can adjust it later.

joevdh: I'm not at all opposed to what you suggest!  I just don't see how it will work without also modifying the constructor for the base class (LLGL::Texture) so its no longer RAII.  Happy to make that change if you think its the right way to go.

LukasBanana: I see what you mean. You could write a wrapper for both values that return either of them. That means this constructor would invoke `GetDesc()` twice, so if you think that's unnecessary function calls, feel free to keep a boiled down version of the adoption parameters. Maybe move the `bindFlags` and `type` out into the constructor itself, I don't think this struct is necessary (perhaps for Vulkan, but not D3D).

include/LLGL/XR/XRSystemFlags.h
```diff
@@ -0,0 +1,224 @@
...
+\brief Per-view configuration entry returned by XRSystem::GetViewConfigurations.
+\see XRSystem::GetViewConfigurations
+*/
+struct XRViewConfigurationView
+{
+    //! Recommended image size (in pixels) for this view's swap-chain.
+    LLGL::Extent2D recommendedImageExtent = { 0, 0 };
+
+    //! Maximum supported image width (in pixels) for this view's swap-chain.
+    std::uint32_t   maxImageWidth           = 0;
```
LukasBanana: Please don't forget to merge `maxImageWidth` and `maxImageHeight` to also use `Extent2D`.

sources/Renderer/DebugLayer/DbgRenderSystem.cpp
```diff
@@ -433,7 +433,11 @@ RenderTarget* DbgRenderSystem::CreateRenderTarget(const RenderTargetDescriptor&
             {
                 if (debugger_)
                     ValidateAttachmentDesc(attachmentDesc, colorTarget, isResolveAttachment, isDepthStencilAttachment);
-                attachmentDesc.texture = DbgGetInstance<DbgTexture>(attachmentDesc.texture);
+                // Unwrap a DbgTexture to its underlying instance; leave externally-owned textures
+                // (e.g. XR runtime swap-chain images adopted by the XR binding) as-is so the inner
+                // renderer receives a handle it can cast back to its native texture type.
+                if (DbgTexture* textureDbg = dynamic_cast<DbgTexture*>(attachmentDesc.texture))
```
LukasBanana, line 439: This is very hacky as the Debug layer always expects these wrappers. Where does the XR system need to cast it to an underlying type? The XR system should instead check whether it needs to cast it to a `DbgTexture` and then access the underlying texture, but the Debug layer should not be pried open for the XR system. The XR system could also use `GetNativeHandle()` if it needs to access the underlying texture resources.

joevdh: Its not that the XR system needs to cast to the underlying type.  Its that OpenXR insists on creating the images itself rather than going through the usual LLGL image creation flow.  These images then need to be transferred to the render-api specific LLGL wrapper.  The call stack goes from OpenXRSession::CreateSwapChain() -> <RenderBackend>OpenXRGraphicsBinding::EnumerateSwapchainImages().  It doesn't create the images through the RenderAPI wrapper, so the debug layer never gets a chance to intercept the call and wrap the result in a DbgTexture.  So when the image is used later, the debug layer in the old code assumed that the image is a DbgTexture and casted it without checking.  

To make this work as the debug RENDER layer expects, we'd need to add an XR equivalent of a the Render DebugLayer, which was more work I wasn't looking to bite off, considering this PR is already massive.  Considering that DbgRenderSystem::ValidateAttachmentDesc wasn't actually accessing any data that wasn't already exposed by the base Texture class, the lesser evil seemed to be to just unwrap the unnecessary debug texture (if any) and forward it on.

include/LLGL/XR/XRSystemFlags.h
joevdh: Could have sworn I did that!  Maybe it was in another struct.

sources/Renderer/Direct3D11/Texture/D3D11Texture.h
joevdh: Ok, change made.  D3D11 & 12 take just the necessary args as contructor params, and Vulkan has the struct moved out to the LLGL name space.  Thoughts?

sources/Renderer/DebugLayer/DbgRenderSystem.cpp
LukasBanana: I think this reveals a wider issue with the fact that the ownership of those textures does not fit into how LLGL manages its resources. A user could call `Release()` on them but that's not allowed anymore because they are not owned by `RenderSystem` and it doesn't work with the Debug layer as this requires them to be wrapped differently, which also applies to all functions that take a `Texture&` parameter.

Are these textures only ever needed to be attached to a render target so that the runtime like in HelloOpenXR can render into it during a render-pass, or could these textures returned by your `GetImages()` function ever be used for custom sampling or bound to a PSO (e.g. to write to them in a compute shader)? I assume they are only intended for this very specific render-target since using them for any pipeline layout that defines these resources with properties they were not created with (e.g. RW access, UAV/SRV properties etc.) would result in undefined behavior. If this is the case, we should completely hide these textures from the user and instead provide a `GetRenderTargets()` function that returns these render-targets predefined for the `XRSwapChain`. That doesn't fully solve the problem (yet) but it reduces the amount of access we're exposing to the public interface, because there is not much you can do with a `RenderTarget` other than using it for a render-pass section. A `Texture` on the other hand can be bound to any PSO, which these should not be used for as I understand it. We also don't have to worry about the Debug wrapper in functions like `WriteTexture()`/`ReadTexture()` etc. The only function for `RenderTarget` we have to modify in order to detect invalid usage is `Release()`.

Sorry to make you revise this a couple of times, but I think it's important we get this right. We need to be strict about what system owns the resources and I think by only exposing the `RenderTarget` interface, it is easier to limit the possibility of invalid use.
For the long-term solution, we can come up with an interface similar to the native render system you pass into `RenderSystem::Load()`, so you can create externally owned textures and buffers.

joevdh: Ok, so just that I'm clear on the changes you're asking for, it sounds like basically you want to move the majority of the code in MyXRRenderer::CreateSwapChains() into the LLGL XR code.  So LLGL will do all render target and depth image setup internally, so that the Textures used by XRSwapBuffer never have to be exposed.  Is that right?

LukasBanana: Yes, I think it's better if we can hide the XR textures as much as we can and let the XR code handle the render target creation. This won't fix the debug layer issue, but I think I can handle that better if we only expose the `RenderTarget` interface instead of the `Texture` interface since those are intertwined with more functions and other interfaces. Perhaps you can find a way to integrate the `DbgRenderTarget` wrapper but if not then I will try to take care of that once your PR went in.

joevdh: Ok, moved all that inside of the LLGL code, no longer exposing it to the user.  The API actually looks more like the non-LLGL path now

BuildAndroid.sh
```diff
@@ -277,13 +308,22 @@ generate_app_project()
     mkdir -p "$APP_ROOT/gradle/wrapper"
     cp -r "$PLATFORM_SOURCE_DIR/gradle/wrapper/gradle-wrapper.properties" "$APP_ROOT/gradle/wrapper/gradle-wrapper.properties"
 
+    # XR projects mark themselves with an Android.openxr file. When present, swap in the XR
+    # manifest variant (immersive_hmd feature, OPENXR permissions, broker queries, intent
+    # category) and remember to bundle the loader .so below.
+    NEEDS_OPENXR_LOADER=0
+    if [ -f "$PROJECT_SOURCE_DIR/Android.openxr" ]; then
+        NEEDS_OPENXR_LOADER=1
+        cp "$PLATFORM_SOURCE_DIR/app-xr/src/main/AndroidManifest.xml" "$APP_ROOT/app/src/main/AndroidManifest.xml"
```
LukasBanana, line 317: Is the file *app/src/main/AndroidManifest.xr.xml* still needed since you now have a separate folder and just copy the correct one into the destination folder?

sources/Renderer/Vulkan/Ext/VKExtensionRegistry.cpp
```diff
@@ -17,54 +18,78 @@ namespace LLGL
 
 static bool g_VKRegisteredExtensions[static_cast<std::size_t>(VKExt::Count)] = {};
 
-static const char* g_VKOptionalExtensions[] =
-{
-    #if VK_EXT_conditional_rendering
-    VK_EXT_CONDITIONAL_RENDERING_EXTENSION_NAME,
+static const char *g_VKOptionalExtensions[] =
```
LukasBanana: Would you mind reverting the formatting for this entire initializer list? I'm trying reduce the size of this PR a little bit and this here looks like auto-formatted. This should only have the lines that add new entries (and their comments of course).

LukasBanana (commented):
I'm back from vacation, thanks for your patience. I only left two more notes in an attempt to reduce the size of this massive PR slightly. I think the `RenderTarget` interface works better from an API standpoint for the reasons we discussed. If you could take a look at these notes, I'm happy to finally merge this into the main stream and then take a stab at how we can improve the situation with the debug layer.

BuildAndroid.sh
joevdh: Good catch.  I'll remove it

joevdh:
Changes made!

sources/Renderer/Vulkan/Ext/VKExtensionRegistry.cpp
```diff
@@ -18,53 +19,77 @@ namespace LLGL
 static bool g_VKRegisteredExtensions[static_cast<std::size_t>(VKExt::Count)] = {};
 
 static const char* g_VKOptionalExtensions[] =
-{
-    #if VK_EXT_conditional_rendering
-    VK_EXT_CONDITIONAL_RENDERING_EXTENSION_NAME,
+    {
```
LukasBanana: @joevdh I actually meant the entire file, all those indentations that seem unnecessary in this PR.

LukasBanana (approved):
Thanks again for this great contribution ❤️ 

There might be an issue with the UWP runners, but I'll ignore them as they seem unrelated to the PR. Once the remaining succeed, I'll merge them.

LukasBanana:
@joevdh heads up that 07641bc reverts `DbgRenderSystem::ValidateAttachmentDesc()` back to using DbgTexture descriptor instead of `Texture::GetDesc()` for two reasons:
1. We want to avoid `dynamic_cast` so that LLGL can be used in a strictly RTTI free build; only _CheckedCast.h_ should use it when `LLGL_ENABLE_CHECKED_CAST` is enabled.
2. The validation layer is meant to validate the DbgTexture descriptor since `Texture::GetDesc()` is allowed to return attributes that deviate from the input, e.g. when you request a texture format like RGB, the backends are allowed to change it to RGBA if RGB is not supported natively and the backend converts it on the fly.

This shouldn't have any impact on your XR work.

LukasBanana:
After some tinkering with my new PSVR2 adpater for PC, I finally got the **HelloOpenXR** example to work and added a screenshot to the [example page](https://github.com/LukasBanana/LLGL/tree/master/examples/Cpp#helloopenxr) 🍾 

The call to `xrGetInstanceProcAddr(instance, "xrGetD3D12GraphicsRequirementsKHR", ...)` would return `XR_ERROR_HANDLE_INVALID` and report this in the log:
```
Error [GENERAL | xrGetInstanceProcAddr | OpenXR-Loader] : No active XrInstance handle.
```
Once I switched to **static build**, the OpenXR loader worked fine, so something with the dynamic loader might be wrong on my end.

FWIW: After some issues with my PSVR2 controllers, I figured that it was time to finally install the WiFi antenna with my motherboard; I didn't realize they were also helpful for Bluetooth. So that was a good exercise 😄 

@joevdh Let me know if you'd like to add your name anywhere in the code, I encourage people that add an entirely new feature to add their name in the doxygen docu via `\author{...}`, but no pressure.

joevdh:
Shoot.  The OpenXR loader is a challenge to work with, sadly.  The new multiview PR has some fixes to it, building OpenXR as a DLL when LLGL builds DLLs and and linking it statically when LLGL is linked statically.  Can you try both options on the new branch, and if either they don't work LMK?

PR #236, opened 2026-05-14
Add multiview support

joevdh:
Added support for multiview rendering, usable for both XR and non-XR use cases (eg: cascading shadow maps).


joevdh:
There's still problems with this implementation.  Switching to draft

PR #237, opened 2026-06-02
Fix Linux build with Wayland enabled + Some changes

st0rmbtw:


sources/Platform/Linux/Wayland/LinuxWaylandState.cpp
```diff
@@ -139,6 +145,19 @@ void LinuxWaylandState::HandleRegistryGlobal(
...
+    LinuxWaylandState& state = GetInstance();
+
+    // Check if the name corresponds to any wl_output, if so destroy it
+    for (auto it = state.displayList_.begin(); it != state.displayList_.end(); ++it)
+    {
+        LinuxDisplayWayland* display = *it;
+
+        if (name == display->GetData().name)
+        {
+            delete display;
```
LukasBanana: Can we make this an `std::unique_ptr` instead of manual `new` and `delete`? The only place where I'd put a `delete` is in a destructor for the "pimpl" idiom where it's obvious it's only called once and there's always a corresponding `new` in the constructor. Any other kind of manual delete is unsafe and I try to avoid it entirely in any C++11 project.

sources/Platform/Linux/Wayland/LinuxWaylandState.cpp
```diff
@@ -653,13 +696,38 @@ void LinuxWaylandState::RemoveWindow(LinuxWindowWayland *window)
     }
 }
 
+void LinuxWaylandState::RemoveDisplay(LinuxDisplayWayland* display)
+{
+    LinuxWaylandState& instance = GetInstance();
+
+    for (auto it = instance.displayList_.begin(); it != instance.displayList_.end(); ++it)
```
LukasBanana: This looks like it can be replaced with `LLGL::RemoveFromListIf()` from *CoreUtils.h*.

LukasBanana (changes requested):

sources/Platform/Linux/Wayland/LinuxDisplayWayland.cpp
```diff
@@ -20,9 +20,19 @@ LinuxDisplayWayland::LinuxDisplayWayland(const WaylandDisplayData& data) :
...
+    {
+        wl_output_destroy(data_.output);
+        data_.output = nullptr;
+    }
+}
+
 bool LinuxDisplayWayland::IsPrimary() const
 {
-    return (this == LinuxWaylandState::GetDisplayList()[0]);
+    return (this == LinuxWaylandState::GetDisplayList()[0].get());
```
st0rmbtw, line 35: Would it be better if we identified the display by its id (`data_.name`) instead of by the pointer?

LukasBanana: I'm not familiar with Wayland. Gemini suggests Wayland leaves the concept of a primary display up to the compositors and that a heuristic might be necessary like checking the position being at `(0, 0)`. What name would you be looking for to detect the primary display?

sources/Platform/Linux/LinuxDisplay.cpp
```diff
@@ -162,6 +161,9 @@ Offset2D Display::GetCursorPosition()
     return Offset2D{ 0, 0 };
 }
 
+void LinuxDisplay::ClearCache()
+{
```
LukasBanana, line 165: Can you please add a comment to empty functions that says "dummy" or an explanation why it's empty. This makes it clear they are empty on purpose.

LukasBanana (changes requested):
Looks good. I'll merge if you can please add a comment to the empty function body.

sources/Platform/Linux/Wayland/LinuxDisplayWayland.cpp
st0rmbtw: What I meant was, should we instead of comparing the pointers
```cpp
this == LinuxWaylandState::GetDisplayList()[0].get()
```
compare the ids 
```cpp
data_.name == LinuxWaylandState::GetDisplayList()[0]->data_.name
```

It's not a big deal, just a code quality nit-pick

LukasBanana: Oh I see. Makes sense to me but I leave it up to you, since you know Wayland better than me.

LukasBanana (approved):

PR #238, opened 2026-06-25
Add ARM64 build support for Windows

joevdh:
Detect ARM64 in CMake via CMAKE_GENERATOR_PLATFORM / CMAKE_SYSTEM_PROCESSOR and set the 64-bit and architecture flags accordingly. Add a single-config (e.g. Ninja) fallback through CMAKE_SIZEOF_VOID_P, define LLGL_ARCH_ARM64 in Platform.h, and let BuildWin64.bat honor a pre-set BUILD_ARCH (e.g. ARM64) instead of hard-coding x64.

### DIFF
--- a/BuildWin64.bat
+++ b/BuildWin64.bat
@@ -3,7 +3,8 @@
 set SOURCE_DIR=%~dp0
 set OUTPUT_DIR=build_win64
 set BUILD_TYPE=Release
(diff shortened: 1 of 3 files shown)


LukasBanana (approved):
Looks good to me 👍 Thank you

PR #240, opened 2026-06-25
Vulkan: recreate swap-chain on out-of-date instead of trapping

joevdh:
vkAcquireNextImageKHR / vkQueuePresentKHR can report VK_ERROR_OUT_OF_DATE_KHR or VK_SUBOPTIMAL_KHR when the surface no longer matches the swap-chain (window resize, DPI change, minimized window, or first present on some drivers such as Adreno). Previously these were treated as fatal.

Recreate the surface and resolution-dependent resources and skip the frame instead. A swapChainOutOfDate_ flag defers recreation while the surface is zero-area (e.g. a minimized window) and retries it on the next Present().

### DIFF
--- a/sources/Renderer/Vulkan/VKSwapChain.cpp
+++ b/sources/Renderer/Vulkan/VKSwapChain.cpp
@@ -99,6 +99,16 @@ bool VKSwapChain::IsPresentable() const
 
 void VKSwapChain::Present()
 {
+    /*
+    If a previous Present/Acquire reported the swap-chain as out-of-date but it could not be
+    recreated yet (e.g. a zero-size/minimized window), retry the recreation and skip this frame.
+    */
+    if (swapChainOutOfDate_)
+    {
+        RecreateSwapChain();
+        return;
+    }
+
     /* Initialize semaphores */
     VkPipelineStageFlags waitStages[] = { VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT };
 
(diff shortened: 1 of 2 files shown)


sources/Renderer/Vulkan/VKSwapChain.cpp
```diff
@@ -131,6 +141,18 @@ void VKSwapChain::Present()
...
     }
     result = vkQueuePresentKHR(presentQueue_, &presentInfo);
+
+    /*
+    VK_ERROR_OUT_OF_DATE_KHR / VK_SUBOPTIMAL_KHR are not fatal: the surface no longer matches the
+    swap-chain (window resize, DPI change, or the first present on some drivers, e.g. Adreno).
+    Recreate the swap-chain instead of trapping. RecreateSwapChain() re-acquires the next image,
+    so we return here rather than falling through to AcquireNextColorBuffer().
+    */
+    if (result == VK_ERROR_OUT_OF_DATE_KHR || result == VK_SUBOPTIMAL_KHR)
```
LukasBanana, line 151: I somehow missed this from [this Vulkan tutorial](https://vulkan-tutorial.com/Drawing_a_triangle/Swap_chain_recreation), but they also state those are considered success codes. Naming is hard :)

LukasBanana (approved):

LukasBanana:
I tried to go back to Merged Commit on GitHub, but I do not like the commit noise. So I force pushed it back to Squashed Commits. Sorry for the interruption.

PR #241, opened 2026-06-26
Handle the `StorageBuffer` storage class introduced in SPIR-V 1.3

st0rmbtw:
Since version 1.3, SPIR-V has a separate storage class for SSBOs

### DIFF
--- a/sources/Renderer/SPIRV/SpirvReflect.cpp
+++ b/sources/Renderer/SPIRV/SpirvReflect.cpp
@@ -413,6 +413,7 @@ SpirvResult SpirvReflect::OpVariable(const Instr& instr)
     {
         case spv::StorageClassUniform:
         case spv::StorageClassUniformConstant:
+        case spv::StorageClassStorageBuffer:
         {
             auto& var = uniforms_[instr.result];
             {


sources/Renderer/SPIRV/SpirvReflect.cpp
```diff
@@ -413,6 +413,7 @@ SpirvResult SpirvReflect::OpVariable(const Instr& instr)
     {
         case spv::StorageClassUniform:
         case spv::StorageClassUniformConstant:
+        case spv::StorageClassStorageBuffer:
```
LukasBanana, line 416: Can you please show me a brief GLSL example where this new storage class would be used? This switch-case allocates new _uniform_ variables, so this storage class doesn't immediately look appropriate here. Is this storage class related to SSBOs or is it a new concept?

st0rmbtw: I generate GLSL shaders from SPIR-V using `spirv_cross` and SPIR-V itself is generated using `Slang` from its source code.

In the generated GLSL the `StorageBuffer` binding looks like this:
```glsl
layout(binding = 5, std430) readonly buffer TransformBuffer
{
    vec4 _member0[];
} TransformBuffer_1;
```
It worked fine before, but after I updated Vulkan toolchain it broke. If I understand correctly, since SPIR-V 1.3 it emits a different storage class for SSBOs.

There is more info here: https://github.com/KhronosGroup/Vulkan-Guide/blob/main/chapters/extensions/shader_features.adoc#vk_khr_storage_buffer_storage_class

st0rmbtw: After I updated the Vulkan toolchain, my program was crashing with this error:
```
Failed to create a shader. Error: SPIR-V binding table failure: resource '(null)' (set=0, binding=5) could not be resolved to a valid descriptor type
```

I found out that it's because `TransformBuffer` binding wasn't being reflected because it's storage class was `StorageBuffer` and LLGL didn't handle it.

LukasBanana (approved):
Thanks for the detailed information. So the KhronosGroup made a breaking change and LLGL needs to handle it now, got it.

PR #242, opened 2026-06-26
Added multiview (single-pass stereo) rendering for Vulkan and Direct3…

joevdh:
Added multiview (single-pass stereo) rendering for Vulkan and Direct3D 12.

API is generic so can be used in multiple different cases (stereo eyes, shadow cascades, etc.), with OpenXR as the first consumer.

- Added SystemValue::ViewIndex (gl_ViewIndex / SV_ViewID)
- Added view count on RenderPassDescriptor and RenderTargetDescriptor
- Added RenderingFeatures::hasMultiView and RenderingLimits::maxViews
- Regenerated C99/C#/Go wrappers and C99 type assertions
- Vulkan: VK_KHR_multiview view mask in the render pass, layered render-target image views, and caps from the multiview extension/properties
- Direct3D 12: view instancing via SV_ViewID, gated on LLGL_D3D12_ENABLE_DXCOMPILER (Shader Model 6.1)
- Updated HelloOpenXR example
- Added unit test

OpenXR fixes uncovered along the way:
- Linking OpenXR statically was causing multiple copies of the same symbol to show up in each of the LLGL dlls when LLGL was not itself statically linked.  Now OpenXR is a DLL unless LLGL is built as a static lib
- Populate device extension names when adopting an XR-created Vulkan device, so extension-gated caps (including multiview) are reported correctly
- SteamVR has a lot of Vulkan validation errors, so rather than whitelist each one (which is a moving target) we only assert on errors that come from LLGL
- Acquire XR swap-chain images on demand inside the frame instead of ahead of time, keeping the acquire within the begin/end-frame bracket. Fixes a head-locked / frozen image on runtimes that reject out-of-bracket acquires with XR_ERROR_CALL_ORDER_INVALID (e.g. VIVE WAVE). XRSwapChain::GetRenderTarget is split into AcquireRenderTarget() (side-effecting) and GetRenderTarget() const.


joevdh:
Sorry, another large PR.  But most files only have a single line change

sources/Renderer/Vulkan/RenderState/VKRenderPass.cpp
```diff
@@ -268,11 +270,34 @@ void VKRenderPass::CreateVkRenderPassWithDescriptors(
         subpassDep.dependencyFlags  = 0;
     }
 
+    /*
+    For multiview (single-pass layered) rendering, broadcast the single subpass to 'numViews_' views via a
+    view mask. The correlation mask hints that all views are rendered from correlated geometry, which lets
+    drivers optimize. This requires VK_KHR_multiview (core in Vulkan 1.1).
+    */
+    const void* createInfoNext = nullptr;
+    #if VK_KHR_multiview
```
LukasBanana, line 280: This will need a runtime check like in [`VKCommandBuffer::DrawMesh`](https://github.com/LukasBanana/LLGL/blob/master/sources/Renderer/Vulkan/Command/VKCommandBuffer.cpp#L1364) and not just a preprocessor check. If you compile this with an SDK that supports Vulkan 1.1 but run it on a machine that doesn't have that extension, I presume it would be undefined behavior or just crash.
When the extension is not supported (and it should be an optional extension) we of course cannot support a view count greater than 1, so we still have to report an error.

include/LLGL/RenderSystemFlags.h
```diff
@@ -772,6 +784,15 @@ struct RenderingLimits
...
     std::uint32_t   maxViewportSize[2]                  = { 0, 0 };
 
+    /**
+    \brief Specifies the maximum number of views for multiview (single-pass layered) rendering. By default 0.
+    \remarks If multiview is not supported, this is 0. Otherwise it is at least 2.
+    This is the upper limit for RenderPassDescriptor::views.
+    \see RenderPassDescriptor::views
+    \see RenderingFeatures::hasMultiView
+    */
+    std::uint32_t   maxViews                            = 0;
```
LukasBanana: If the default value for `RenderTargetDescriptor::views` is 1, wouldn't it make sense to define the minimum required value for `maxViews` to be 1 as well? I'm fine with keeping this default to 0, which indicates this value hasn't been initialized by the renderer yet, but I'm referring to the contract in the comment above _"If multiview is not supported, this is 0"_. Should we expect the renderer to return this as 0 or 1 if multiview is not supported?

sources/Renderer/Direct3D12/RenderState/D3D12GraphicsPSO.cpp
```diff
@@ -241,6 +265,126 @@ ComPtr<ID3D12PipelineState> D3D12GraphicsPSO::CreateNativePSOWithDesc(ID3D12Devi
     return pipelineState;
 }
 
+#ifdef LLGL_D3D12_ENABLE_DXCOMPILER
+
+template <D3D12_PIPELINE_STATE_SUBOBJECT_TYPE TType, typename TObject>
+struct alignas(void*) D3DPipelineStreamSubobject
```
LukasBanana: I see you adopted this from _D3D12MeshPSO.cpp_. Can we please move this template into a new file like _D3D12PSOUtils.h_ to consolidate them?

sources/Renderer/Direct3D12/D3D12RenderSystem.cpp
```diff
@@ -876,6 +876,16 @@ void D3D12RenderSystem::QueryRenderingCaps(RenderingCapabilities& caps)
     caps.features.hasOffsetInstancing               = (featureLevel >= D3D_FEATURE_LEVEL_9_3);
     caps.features.hasIndirectDrawing                = (featureLevel >= D3D_FEATURE_LEVEL_10_0);//???
     caps.features.hasViewportArrays                 = true;
+    #ifdef LLGL_D3D12_ENABLE_DXCOMPILER
```
LukasBanana: This macro is only relevant to enable the HLSL source input for SM6+. Even without it, you should still be able to feed DXIL into LLGL if it the HLSL was compiled offline, so we don't need to gate the multiview feature behind this macro.

sources/Renderer/OpenGL/Shader/GLShaderProgram.cpp
```diff
@@ -418,6 +418,7 @@ static SystemValue FindSystemValue(const StringView& name)
         { "gl_SampleID",        SystemValue::SampleID           },
         { "gl_VertexID",        SystemValue::VertexID           }, // GLSL
         { "gl_VertexIndex",     SystemValue::VertexID           }, // SPIR-V
+        { "gl_ViewIndex",       SystemValue::ViewIndex          },
```
LukasBanana, line 421: I presume this is just for the sake of completion.

LukasBanana (commented):
I only skimmed through it and left a few comments, but I will take a closer look over the weekend. It looks like another welcome addition :)

sources/Renderer/Vulkan/RenderState/VKRenderPass.cpp
joevdh: Good call.

include/LLGL/RenderSystemFlags.h
joevdh: Yeah, good point.  The user is already getting the information about whether multiview is supported from hasMultiView, this is just redundant and will make code more tricky.

sources/Renderer/Direct3D12/D3D12RenderSystem.cpp
joevdh: I believe in the example projects (and probably the apps consuming LLGL) the HLSL shaders are compiled at runtime by the app, so the user could in theory do everything right to use multiview, but because the shaders are compiled with FXC instead of DXC it won't work and won't be very clear why.  I agree its not _strictly_ necessary, but its seems like a reasonable requirement to expect SM6.1 support across the pipeline if you want to use the feature.

That said, I'm happy to make the change if that's what you'd prefer.

sources/Renderer/OpenGL/Shader/GLShaderProgram.cpp
joevdh: Well, inserting ViewIndex in SystemValue bumped the value of ViewportIndex, so the entries in this array wouldn't line up with the enum.

sources/Renderer/Direct3D12/D3D12RenderSystem.cpp
LukasBanana: Yes, the examples compile it at runtime and most of the other examples don't use SM6 shaders. The "MeshShaders" unit test uses SM6.5 shaders and blindly assumes DXC is included. So this needs improvement and we also need entries for DXIL in the `ShadingLanguage` enum so if DXC is not included, `HLSL_6_0` and up would not be included but `DXIL` would still be included, so the runtime can check whether they need to be loaded as binaries only. Let me open a bug ticket for that.
As for multiviews, I think we should not make it dependent on this option, since we can in fact use it without DXC attached to the library.

tests/Testbed/Shaders/CompileAllGlslShadersToSpirv.bat
```diff
@@ -39,4 +39,6 @@ glslangValidator -V --target-env vulkan1.2 -DENABLE_SPIRV=1 -o Meshlet\Meshlet.4
 glslangValidator -V -DENABLE_SPIRV=1 -o Meshlet\Meshlet.450core.frag.spv Meshlet\Meshlet.450core.frag
 
 echo DONE
+glslangValidator -V -DENABLE_SPIRV=1 -o Multiview\Multiview.450core.vert.spv Multiview\Multiview.450core.vert
```
LukasBanana, line 40: This looks like it slipped under the `DONE` output.

tests/Testbed/UnitTests/TestMultiview.cpp
```diff
@@ -0,0 +1,153 @@
...
+    auto HasLanguage = [this](ShadingLanguage lang) -> bool
+    {
+        return (std::find(caps.shadingLanguages.begin(), caps.shadingLanguages.end(), lang) != caps.shadingLanguages.end());
+    };
+
+    Shader* vertShader = nullptr;
+    Shader* fragShader = nullptr;
+    if (HasLanguage(ShadingLanguage::HLSL))
+    {
+        vertShader = LoadShaderFromFile("Multiview.hlsl", ShaderType::Vertex,   "VSMain", "vs_6_1", nullptr, VertFmtEmpty);
```
LukasBanana, line 62: Thanks for adding a unit test for this new feature. I'm fine with keeping the loading of such shaders in the tests themselves, perhaps it's not a bad idea to move some of the globally provided shaders (i.e. in `TestbedContext::LoadShaders()`) to their respective unit tests afterall. FWIW this was originally intended for shaders that are reused by multiple tests, but most of them use their own unique shaders anyway.

sources/Renderer/Direct3D12/RenderState/D3D12GraphicsPSO.cpp
```diff
@@ -241,6 +265,112 @@ ComPtr<ID3D12PipelineState> D3D12GraphicsPSO::CreateNativePSOWithDesc(ID3D12Devi
     return pipelineState;
 }
 
+#ifdef LLGL_D3D12_ENABLE_DXCOMPILER
```
LukasBanana: As discussed, we should not make this dependent on the DXC compiler setting.

Instead, this block needs to be wrapped into `#if LLGL_D3D12_ENABLE_FEATURELEVEL >= 1` just like _D3D12MeshPSO.cpp_ to exclude it from builds that don't have the latest D3D12 SDK; the MSVC14 doesn't support it for instance, that's why the [AppVeyor CI builds](https://ci.appveyor.com/project/LukasBanana/llgl/builds/54286714/job/x3g3hcya6o81c6qw#L343) are failing in _D3D12PSOUtils.h_. They don't fail on this particular file because those builds don't have DXC enabled.

The name "feature level" might be misused here and I'm open for suggestions how to improve it as that name is conflated within LLGL between SDK versions, D3D feature levels, and D3D versions.

sources/Renderer/Direct3D12/RenderState/D3D12PSOUtils.h
```diff
@@ -0,0 +1,55 @@
+/*
+ * D3D12PSOUtils.h
+ *
+ * Copyright (c) 2015 Lukas Hermanns. All rights reserved.
+ * Licensed under the terms of the BSD 3-Clause license (see LICENSE.txt).
+ */
+
+#ifndef LLGL_D3D12_PSO_UTILS_H
+#define LLGL_D3D12_PSO_UTILS_H
+
```
LukasBanana, line 10: `#if LLGL_D3D12_ENABLE_FEATURELEVEL >= 1` needs to be inserted here, because `D3D12_PIPELINE_STATE_SUBOBJECT_TYPE` is not defined with older D3D12 SDKs. That's why the [AppVeyor CI builds](https://ci.appveyor.com/project/LukasBanana/llgl/builds/54286714/job/x3g3hcya6o81c6qw#L343) are failing.

sources/Renderer/Direct3D12/RenderState/D3D12PSOUtils.h
```diff
@@ -0,0 +1,55 @@
...
+    D3D12_PIPELINE_STATE_SUBOBJECT_TYPE type = {TType};
+    TObject object = {};
+
+    D3DPipelineStreamSubobject() = default;
+
+    inline D3DPipelineStreamSubobject(TObject object) : object{std::move(object)}
+    {
+    }
+
+    inline D3DPipelineStreamSubobject &operator=(TObject object)
```
LukasBanana: Minor note, you seem to have auto-formatting enabled when you paste code. You can usually prevent this by pasting and then one time undo should only revert the auto-formatting. This way you can move the code as-is.

include/LLGL/RenderTargetFlags.h
```diff
@@ -139,6 +139,23 @@ struct RenderTargetDescriptor
     */
     std::uint32_t           samples     = 1;
 
+    /**
+    \brief Specifies the number of views for multiview (single-pass layered) rendering. By default 1.
```
LukasBanana, line 143: This contract should also be implemented in the debug layer. In `DbgRenderSystem::CreateRenderTarget()` we should add something like this (maybe in a new `Validate...` function):
```cpp
/* Get number of views from render-target descriptor or render-pass and warn on mismatch */
std::uint32_t numViews = renderTargetDesc.views;
if (renderTargetDesc.renderPass != nullptr)
{
    const DbgRenderPass* renderPassDbg = LLGL_CAST(const DbgRenderPass*, renderTargetDesc.renderPass);
    numViews = renderPassDbg->desc.views;
    if (renderTargetDesc.views > 1 && renderPassDbg->desc.views > 1 && renderTargetDesc.views != renderPassDbg->desc.views)
    {
        LLGL_DBG_WARN(
            WarningType::ImproperArgument,
            "mismatching number of views between render-target (%u) and render-pass (%u)",
            renderTargetDesc.views, renderPassDbg->desc.views
        );
    }
}

/* Validatge multiview feature support */
if (numViews > 1)
{
    if (!GetRenderingCaps().features.hasMultiview)
    {
        LLGL_DBG_ERROR(
            ErrorType::UnsupportedFeature,
            "multiview not supported, but render-target specified %u",
            renderTargetDesc.views
        );
    }
    else if (numViews > GetRenderingCaps().limits.maxViews)
    {
        LLGL_DBG_ERROR(
            ErrorType::InvalidArgument,
            "maximum number of supported views is %u, but render-target specified %u",
            GetRenderingCaps().limits.maxViews, numViews
        );
    }
}

```

sources/Renderer/Direct3D12/Texture/D3D12RenderTarget.cpp
```diff
@@ -471,24 +507,25 @@ void D3D12RenderTarget::CreateDepthStencilView(
...
             dsvDesc.Texture2DArray.FirstArraySlice      = arrayLayer;
-            dsvDesc.Texture2DArray.ArraySize            = 1;
+            dsvDesc.Texture2DArray.ArraySize            = numArrayLayers;
             break;
 
         case TextureType::Texture1DArray:
             dsvDesc.ViewDimension                       = D3D12_DSV_DIMENSION_TEXTURE1DARRAY;
             dsvDesc.Texture1DArray.MipSlice             = mipLevel;
             dsvDesc.Texture1DArray.FirstArraySlice      = arrayLayer;
-            dsvDesc.Texture1DArray.ArraySize            = mipLevel;
```
LukasBanana, line 481: Wooops, classic copy'n'paste typo.

sources/Renderer/Vulkan/Command/VKCommandBufferRegistry.cpp
```diff
@@ -0,0 +1,65 @@
...
+    if (!enabled)
+        g_commandBufferRegistry.clear();
+}
+
+void VKRegisterCommandBuffers(const VkCommandBuffer* commandBuffers, std::uint32_t count)
+{
+    if (commandBuffers == nullptr)
+        return;
+    std::lock_guard<std::mutex> guard{ g_commandBufferRegistryMutex };
+    if (!g_commandBufferTrackingEnabled)
```
LukasBanana, line 37: I would make `g_commandBufferTrackingEnabled` an atomic boolean to avoid having to enter the critical section every time.

sources/Renderer/Vulkan/Command/VKCommandBufferRegistry.h
```diff
@@ -0,0 +1,48 @@
...
+the same VkDevice (e.g. an OpenXR runtime compositor, which records its own barriers/draws on our device). The
+registry is only populated while tracking is enabled, which the debug messenger turns on; outside debug usage it
+is inert, so there is no overhead in release builds.
+*/
+
+// Enables or disables command-buffer tracking. The debug messenger enables it when it is installed.
+void VKSetCommandBufferTrackingEnabled(bool enabled);
+
+// Registers command buffers LLGL just allocated. No-op while tracking is disabled.
+void VKRegisterCommandBuffers(const VkCommandBuffer* commandBuffers, std::uint32_t count);
```
LukasBanana: Conventionally, when I don't use `ArrayView`, I put the counter first and call it according to its pointer parameter, e.g. `(numCommandBuffers, commandBuffers)`. Examples are `CommandBuffer::SetVertexBuffer()`/ `SetViewports()`/ `ClearAttachments()`, `RenderSystem::CreateBufferArray()`. For byte sizes, I do the opposite, e.g. `WriteBuffer()`, `SetUniforms()`, `UpdateBuffer()`.

sources/Renderer/Vulkan/VKRenderSystem.cpp
```diff
@@ -70,15 +71,15 @@ VKRenderSystem::VKRenderSystem(const RenderSystemDescriptor& renderSystemDesc) :
         if (isDebugLayerEnabled_)
         {
             // The instance was created externally (e.g. by the OpenXR graphics binding); we can only
-            // install the debug callback if that creator enabled VK_EXT_debug_report. Probe via the
+            // install the debug messenger if that creator enabled VK_EXT_debug_utils. Probe via the
```
LukasBanana, line 74: I'm looking at this to a later hour and I might have overlooked it, but can you please give me some info why we're using the **debug_utils** extension instead of the **debug_report** one? Was this superseeded and is it also supported in the same Vulkan version or does this raise the minimum required Vulkan version for debug callbacks?

LukasBanana (changes requested):
Looks pretty good overall 👍 I left a few comments, mostly about minor changes.

I also opened a bug ticket #243 I will take care off soon(-ish) so we don't have to rely on the DXC option to support multiview as I think it shouldn't be a dependency if you can still use it with pre-compiled shaders.

Beyond your PR, since you mentioned you used the wrapper generator: Do you have any feedback about it? Like was it hard to figure out where to find and how to use it? I think I should probably mentioned it in some README.md document or just in the code or even just in the `static_assert` statements.

Also thanks for adding a new unit test for this new feature ❤️

sources/Renderer/Vulkan/VKRenderSystem.cpp
joevdh: Yeah, debug_report was deprecated in favor of debug_utils.  The motivation for the change is that SteamVR keeps emitting a bunch of warnings when you have validation layers enabled, and LLGL had no way to tell where the errors came from and would assert.  debug_utils provides options to tag objects, so we can tag the objects that LLGL makes so when we get a validation message we can check whether the issues is ours or SteamVRs.  

Apparently its been a know issue with SteamVR for a while but hasn't been fixed.  Guess everyone is using D3D on PC.  <shrug>

Though its a newer extension, it still works with Vulkan 1.0 so doesn't require an upgrade, but as with all extensions the runtime has to support it.  It came out more recently than debug_report (2018 vs 2015), but it seems like its old enough it should be ok to make the validation asserts rely on it, and on really old systems that haven't been update its absence should be non-breaking.

sources/Renderer/Vulkan/Command/VKCommandBufferRegistry.cpp
joevdh: The mutex is necessary for protecting access to g_commandBufferRegistry.  I could make g_commandBufferTrackingEnabled atomic and pull it outside the mutex, but then you risk having a case where g_commandBufferTrackingEnabled gets changed on one thread while g_commandBufferRegistry is getting modified on the other... in the current code they'll always be changed together and won't get out of sync.

LukasBanana: > _[...] then you risk having a case where g_commandBufferTrackingEnabled gets changed on one thread while g_commandBufferRegistry is getting modified on the other_

Not if you keep the order in all other functions, keep the mutex, and duplicate this condition (once at the top and once the critical section is entered). Let's assume this scenario:
1. Thread A reads `g_commandBufferTrackingEnabled == true`.
2. Thread B writes `g_commandBufferTrackingEnabled = false` atomically and enteres the critical section.
3. Thread A continues but has to wait to enter the critical section, because Thread B holds that mutex.
4. Thread B exits the critical section and disabling tracking has completed.
5. Thread A continues and enteres the crtitical section, still assuming tracking is enabled.

So this example shows that with an atomic boolean, we'd have to make this check twice and I've seen this pattern before. The benefit is that while tracking is disabled, we can more efficiently exit this function (avoiding entering and exiting critical sections), but we have to check this value again after we entered the critical section to avoid the issue you mentioned.

sources/Renderer/Vulkan/VKRenderSystem.cpp
LukasBanana: Thanks for the explanation, if it's a non-breaking change and we simply don't support this debug utility on older systems, then I'm fine with this.

tests/Testbed/Shaders/CompileAllGlslShadersToSpirv.bat
joevdh: ( facepalm )

sources/Renderer/Vulkan/Command/VKCommandBufferRegistry.cpp
joevdh: Oh, do a double check!  Ok, gotcha

joevdh:
Ok, I think I covered all the feedback.  Have a look!

sources/Renderer/Vulkan/Command/VKCommandBufferRegistry.cpp
```diff
@@ -0,0 +1,73 @@
...
+        if (commandBuffers[i] != VK_NULL_HANDLE)
+            g_commandBufferRegistry.insert(commandBuffers[i]);
+    }
+}
+
+void VKUnregisterCommandBuffers(std::uint32_t count, const VkCommandBuffer* commandBuffers)
+{
+    if (commandBuffers == nullptr)
+        return;
+    std::lock_guard<std::mutex> guard{ g_commandBufferRegistryMutex };
```
LukasBanana, line 59: This would need the same double check of the atomic, right?

sources/Renderer/DebugLayer/DbgRenderSystem.cpp
```diff
@@ -423,23 +423,23 @@ RenderTarget* DbgRenderSystem::CreateRenderTarget(const RenderTargetDescriptor&
...
     LLGL_DBG_SOURCE();
 
+    ValidateRenderTargetDesc(renderTargetDesc);
+
     RenderTargetDescriptor instanceDesc = renderTargetDesc;
     {
         instanceDesc.renderPass = DbgGetInstance<DbgRenderPass>(renderTargetDesc.renderPass);
 
-        auto TransferDbgAttachment = [this](AttachmentDescriptor& attachmentDesc, std::uint32_t colorTarget, bool isResolveAttachment, bool isDepthStencilAttachment)
+        auto TransferDbgAttachment = [this](AttachmentDescriptor &attachmentDesc, std::uint32_t colorTarget, bool isResolveAttachment, bool isDepthStencilAttachment)
```
LukasBanana: I still see some noise in these PRs due to auto-formatting. It might be time for a _.clang-format_ file. Let me get that ready in a separate change, but in the meantime, would you mind removing those re-formatting changes?

**UPDATE**
I am getting frustrated with clang-format, I do not like that you can't turn off formatting rules entirely, it has to be either one style or another, but I want to keep some freedom of choice for how people style their code and only apply a few rules like the type formatting.
Do you have a way to disable auto-formatting per project?

sources/Renderer/Vulkan/Command/VKCommandBufferRegistry.cpp
```diff
@@ -0,0 +1,73 @@
...
+ * VKCommandBufferRegistry.cpp
+ *
+ * Copyright (c) 2015 Lukas Hermanns. All rights reserved.
+ * Licensed under the terms of the BSD 3-Clause license (see LICENSE.txt).
+ */
+
+#include "VKCommandBufferRegistry.h"
+#include <mutex>
+#include <unordered_set>
+
```
LukasBanana: We need to include `<atomic>` here to fix the [CI build errors on Linux](https://github.com/LukasBanana/LLGL/actions/runs/28558088967/job/84669813858?pr=242#step:5:234).

LukasBanana (changes requested):
I only found two more things to address (one to fix the CI build failure) plus one minor note about re-formatting.
I'll start looking into getting a _.clang-format_ file set up, someone else asked about this previously, too. I'm personally not a huge fan of auto-formatting, but having a CI format checker might help reduce some noise from large PRs.

LukasBanana:
FWIW 6352367 allows checking whether DXC is supported if `HLSL_6_0` or up is included in the list of supported shading languages. Otherwise, the corresponding `DXIL` version is included for the D3D12 backend.

sources/Renderer/DebugLayer/DbgRenderSystem.cpp
joevdh: Hmm, I've typically used visual studio and VSCode, and relied on the .editorconfig to prevent the editor from being 'helpful' and changing the formatting when it shouldn't.  I see LLGL has one, but its pretty empty.  Maybe the easiest path is to fill it out with the coding standard settings for LLGL?  Once that's in, there's tools in the editor to clean up files based on the config.  

If its helpful, I've attached the full editorconfig from my project that uses LLGL.  I use more spacing than you, but its at least a starting point for you to tweak.  
[joes_editorconfig.txt](https://github.com/user-attachments/files/29711096/joes_editorconfig.txt)

sources/Renderer/Vulkan/Command/VKCommandBufferRegistry.cpp
joevdh: (facepalm).  Yes, you're right.  I skipped it before because it wasn't already using g_commandBufferTrackingEnabled

sources/Renderer/DebugLayer/DbgRenderSystem.cpp
LukasBanana: I wasn't aware editorconfig supports this many settings, I might have misinterpreted its limited set of rules from some article at some point. I'm open to expanding it, but I'd like most rules to simply not make changes and keeping code as-is, which is something clang-format doesn't seem to support, it's all or nothing sadly.

sources/Renderer/DebugLayer/DbgRenderSystem.cpp
```diff
@@ -1748,6 +1751,17 @@ void DbgRenderSystem::ValidateResourceHeapDesc(const ResourceHeapDescriptor& res
...
 
+        /*
+        A heap-binding with an array size greater than 1 contributes 'arraySize' descriptors to
+        each descriptor set, not one. The number of resource views must be a multiple of the
+        per-set descriptor count (the sum of the binding array sizes), not of the raw binding
+        count - otherwise valid heaps containing an array binding (e.g. a bindless texture
+        table) are wrongly rejected.
+        */
+        std::size_t numDescriptorsPerSet = 0;
+        for (const auto& binding : bindings)
```
LukasBanana: This would also have to be addressed in `GetNumResourceViewsOrThrow()` [ResourceUtils.cpp]. Can you please move this into a separate PR? It doesn't seem necessary for multiview support.

LukasBanana (changes requested):

LukasBanana (approved):
Thanks again for this great contribution!

LukasBanana:
@joevdh heads up that I renamed `hasMultiView` to `hasMultiview` in e992ca4, to be consistent with the spelling of the Vulkan extension and how it's being used throughout the documentation.

**UPDATE:**
Hositing the "Multiview" test revealed a bug in the stream-based PSO creation, since the WARP renderer seems to struggle with mixing and matching the new and old PSO types. So with 47a31e1, I made it so that LLGL always creates stream-based graphics PSOs when the `ID3D12Device2` is available.
FWIW the WARP renderer is available by adding `--ref` to Testbed to run with the "**ref**erence" implentation of a backend.

PR #245, opened 2026-06-28
Fix RenderDoc crashing on OpenGL backend

### LINKED ISSUE #244: [OpenGL] Impossible to debug with RenderDoc
It's currently impossible to debug LLGL app on the `OpenGL` backend with `RenderDoc`. Whenever the RenderDoc is injected into the app process it crashes.

I've ran into this issue a long time ago, but didn't really have time to investigate it (I thought it was a GPU driver issue). But today I needed to debug my app on OpenGL backend with `RenderDoc` and the issue was still present. So, I decided to investigate it.

I noticed that the app was crashing on the `glDrawArraysInstanced` function call (null pointer derefence inside the GPU driver). Then I noticed that RenderDoc was reporting this error on `glBindBuffer` call:
```
RDOC 021156: [16:36:24]        gl_driver.cpp(3321) - Log     - Got a Debug message from GL_DEBUG_SOURCE_API, type GL_DEBUG_TYPE_ERROR, ID 1282, severity GL_DEBUG_SEVERITY_HIGH:
RDOC 021156: [16:36:24]        gl_driver.cpp(3321) - Log     - 'GL_INVALID_OPERATION error generated. Buffer name does not refer to an buffer object generated by OpenGL.'
RDOC 021156: [16:36:24]  gl_buffer_funcs.cpp( 327) - Error   - Invalid/unrecognised buffer passed: glBindBuffer(GL_ARRAY_BUFFER, 5)
```

According to the error mesasge, RenderDoc can't find buffer with the id `5` inside its internal state. And then I noticed that `LLGL` was creating a new OpenGL context when I created the swap chain:
https://github.com/LukasBanana/LLGL/blob/8ad81041e37eb8d4b8d099239cefe04bbb035540/sources/Renderer/OpenGL/GLSwapChain.cpp#L95

I tried to make it use the existing context instead of creating a new one, so I changed that line to this:
```cpp
context_ = contextMngr.AllocContext(nullptr, /*acceptCompatibleFormat:*/ false, &GetSurface());
```

And it worked. The app didn't crash and there were no errors reported from RenderDoc.

### TL;DR
So, the issue is that, even though the state of the old context is shared with the new context inside OpenGL, RenderDoc doesn't share it's internal data  with the new OpenGL context, therefore it complains on binding the buffer that was created with previous context.

st0rmbtw:
I examined RenderDoc's source code and figured out why it's crashing on switching contexts (#244).

So, the issue occurs in the `WrappedOpenGL::DeleteContext` function:
https://github.com/baldurk/renderdoc/blob/abd06a8d1c5a48bf3a84a628df9d776714215600/renderdoc/driver/gl/gl_driver.cpp#L1109-L1177

What is happening is that RenderDoc hooks into the `wglDeleteContext` function and wraps it with `WrappedOpenGL::DeleteContext`. Inside that function it releases its internal resources allocated for that context and does some other things.

And then still inside that function it deletes OpenGL buffers allocated during the lifetime of the context. And here is the problem: it calls `glDeleteBuffer` function while the new created OpenGL context is set as current, so it actually deletes not the buffers of the  context being deleted, but the ones of the current context (that we set when we created the swap chain).

So, I made a workaround: simply unset the current context before deleting some context. I tested it and it works correctly.

Fixes #244

### DIFF
--- a/sources/Renderer/OpenGL/Platform/Win32/Win32GLContext.cpp
+++ b/sources/Renderer/OpenGL/Platform/Win32/Win32GLContext.cpp
@@ -277,6 +277,7 @@ void Win32GLContext::CreateWGLContext(Surface& surface, Win32GLContext* sharedCo
         {
             if (HGLRC extRenderContext = CreateExplicitWGLContext(hDC_, sharedContext))
             {
+                MakeWGLContextCurrent(hDC_, nullptr);
                 /* Use the extended profile and delete the old standard render context */
                 DeleteWGLContext(hGLRC_);
                 hGLRC_ = extRenderContext;


LukasBanana:
Thanks for tracking this down 👍

LukasBanana:
Are you also filing a bug ticket on RenderDoc or do you think this is primarily a bug in LLGL? While the deleting context must not be active, I haven't seen anything yet that officially states that another context must not be active either when `wglDeleteContext()` is invoked.

st0rmbtw:
Yeah, the OpenGL spec doesn't require unsetting the current context before deleting any other. So, yeah, that's a RenderDoc bug.

I will file a bug ticket for them.

PR #246, opened 2026-07-02
Always set stride to 4 for byte address buffers

st0rmbtw:
My app was crashing when I was using `ByteAddressBuffer`'s with an error from D3D11 indicating that the `NumElements` had an incorrect value. I found out that it was because LLGL was not setting the stride properly.

Currently, LLGL uses the stride of **4** bytes (which is correct for all `ByteAddressBuffer`'s) only for the ones that can be bound as `Storage`. Because it checks for the `D3D11_BUFFER_UAV_FLAG_RAW` flag which is set only if the `Storage` binding flag is set.

In my case the buffer was read-only so it was setting an incorrect stride and because of that `NumElements` was incorrect too.

I'm not an expert in Direct3D, so please correct me if I'm wrong, though the change that I made fixed the issue.

### DIFF
--- a/sources/Renderer/Direct3D11/Buffer/D3D11BufferWithRV.cpp
+++ b/sources/Renderer/Direct3D11/Buffer/D3D11BufferWithRV.cpp
@@ -54,7 +54,7 @@ D3D11BufferWithRV::D3D11BufferWithRV(ID3D11Device* device, const BufferDescripto
     uavFlags_     { GetUAVFlags(desc)         }
 {
     /* Determine stride size either for structured buffers or regular buffers */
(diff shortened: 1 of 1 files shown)


LukasBanana:
Thank you very much, this is spot on 👍 

Since `GetD3DResourceViewFormat()` always returns `DXGI_FORMAT_R32_TYPELESS` for byte address buffers (SRV and UAV) the stride must always be 4, not just for UAVs.

No unit test nor example uses `ByteAddressBuffer` at the moment (I shall come up with one) only the built-in shader to copy textures and buffers uses one, but the D3D11 buffer is created differently than user facing interfaces.

LukasBanana:
Perhaps D3D12 needs a similar change in `D3D12Buffer::CreateGpuBuffer()` or this condition should just move into `GetStorageBufferStride()`. A unit test should give us the answer.

PR #247, opened 2026-07-08
Fix resource heap validation and Null backend for array bindings

joevdh:
A heap binding with arraySize > 1 contributes 'arraySize' descriptors per set, so 'numResourceViews' must be a multiple of the sum of binding array sizes, not the raw binding count. Previously valid heaps containing an array binding (e.g. a bindless texture table) were wrongly rejected.

- Export GetNumExpandedHeapDescriptors() for reuse across backends.
- DbgRenderSystem::ValidateResourceHeapDesc: validate against the expanded descriptors-per-set count, and map each resource view to its binding via GetExpandedHeapDescriptors() instead of a hand-rolled array walk.
- Null backend: count expanded descriptors when computing the descriptor set size.

### DIFF
--- a/sources/Renderer/DebugLayer/DbgRenderSystem.cpp
+++ b/sources/Renderer/DebugLayer/DbgRenderSystem.cpp
@@ -12,6 +12,7 @@
 #include "../TextureUtils.h"
 #include "../CheckedCast.h"
 #include "../RenderTargetUtils.h"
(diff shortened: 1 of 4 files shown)


LukasBanana:
I think we'll have to address the wrong interpretation of descriptors with arrays in various places and add a new unit test for it, but this looks good so far.

PR #248, opened 2026-07-24
Add missing BC6H/BC7 rows to the format attributes table

LukasBanana:
g_formatAttribs is indexed positionally by the Format enum, but the table jumped from BC5SNorm straight to ASTC4x4 while the enum declares BC6HUFloat, BC6HSFloat, BC7UNorm and BC7UNorm_sRGB in between.

Every format at or after ASTC4x4 therefore resolved to the wrong row, so GetFormatAttribs/GetMemoryFootprint reported incorrect block dimensions and bit sizes for all ASTC and ETC formats. BC7 and BC6H themselves were already mapped by every backend (Vulkan, D3D, GL, Metal) and reported by VKPhysicalDevice, so only the shared table was missing.

Table and enum are now 1:1 (115 active rows each).

### DIFF
--- a/sources/Renderer/Format.cpp
+++ b/sources/Renderer/Format.cpp
@@ -170,7 +170,7 @@ static const FormatAttributes g_formatAttribs[] =
     {  64, 4, 4, 1, ImageFormat::Compressed,   DataType::Int8,      Mips | Dim2D_3D | DimCube | Compr | SNorm                  }, // BC4SNorm
     { 128, 4, 4, 2, ImageFormat::Compressed,   DataType::UInt8,     Mips | Dim2D_3D | DimCube | Compr | UNorm                  }, // BC5UNorm
     { 128, 4, 4, 2, ImageFormat::Compressed,   DataType::Int8,      Mips | Dim2D_3D | DimCube | Compr | SNorm                  }, // BC5SNorm
-    { 128, 4, 4, 3, ImageFormat::Compressed,   DataType::Float16,   Mips | Dim2D_3D | DimCube | Compr | SFloat                 }, // BC6HUFloat
+    { 128, 4, 4, 3, ImageFormat::Compressed,   DataType::Float16,   Mips | Dim2D_3D | DimCube | Compr | UFloat                 }, // BC6HUFloat
     { 128, 4, 4, 3, ImageFormat::Compressed,   DataType::Float16,   Mips | Dim2D_3D | DimCube | Compr | SFloat                 }, // BC6HSFloat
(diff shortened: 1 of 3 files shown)


LukasBanana:
@joevdh  I just stumbled across your commit and I took the liberty of creating a PR for this as it's a nasty bug I introduced in f2ea13b7. I'm adding a new unit test for the `Format` enum for this matter.

joevdh:
Oh cool thanks!  I was going to push a PR but just wanted to use it myself a bit first to make sure it didn't introduce anything else.  Discovered a couple more issues that I'll be pushing soon as well

sources/Renderer/Format.cpp
```diff
@@ -170,6 +170,10 @@ static const FormatAttributes g_formatAttribs[] =
     {  64, 4, 4, 1, ImageFormat::Compressed,   DataType::Int8,      Mips | Dim2D_3D | DimCube | Compr | SNorm                  }, // BC4SNorm
     { 128, 4, 4, 2, ImageFormat::Compressed,   DataType::UInt8,     Mips | Dim2D_3D | DimCube | Compr | UNorm                  }, // BC5UNorm
     { 128, 4, 4, 2, ImageFormat::Compressed,   DataType::Int8,      Mips | Dim2D_3D | DimCube | Compr | SNorm                  }, // BC5SNorm
+    { 128, 4, 4, 3, ImageFormat::Compressed,   DataType::Float16,   Mips | Dim2D_3D | DimCube | Compr | SFloat                 }, // BC6HUFloat
```
LukasBanana: @joevdh  Ah, I actually see this one needs to be `UFloat` for `BC6HUFloat`.

joevdh: Good catch!

sources/Renderer/TextureFlags.cpp
```diff
@@ -167,6 +167,34 @@ LLGL_EXPORT Extent3D GetMipExtent(const TextureDescriptor& textureDesc, std::uin
...
+        if (formatAttribs.blockWidth == 0 || formatAttribs.blockHeight == 0)
+            return 0;
+
+        const Extent3D subresourceExtent = CalcTextureExtent(type, extent, subresource.numArrayLayers);
+
+        std::size_t numBlocks = 0;
+        for_range(mipLevel, subresource.numMipLevels)
+        {
+            const Extent3D mipExtent = GetMipExtent(type, subresourceExtent, subresource.baseMipLevel + mipLevel);
+            const std::size_t numBlocksX = (mipExtent.width  + formatAttribs.blockWidth  - 1) / formatAttribs.blockWidth;
```
LukasBanana, line 190: These two lines could use `DivideRoundUp()` from _CoreUtils.h_.

LukasBanana (commented):
Looks good, thanks for addressing this.

PR #249, opened 2026-07-27
Color output msaa

joevdh:
Adds multi-sampling support to OpenXR swap-chains, plus the swap-chain colour format control it depends on.
- `SwapChainDescriptor::format`: request a specific back-buffer format (e.g. an  sRGB one, so the hardware performs the linear→sRGB conversion on write).  `Format::Undefined` keeps the existing auto-pick, so no existing caller changes.
- XR swap-chains honor `sampleCount`. The runtime owns only the final image and it is always single-sampled, so the swap-chain allocates its own multi-sampled attachments and resolves into the acquired image.
- Transient attachments are **inferred rather than requested**: a texture that is  attachment-only, single-MIP, multi-sampled and has no initial data can only be consumed by a resolve inside its own render pass, so Vulkan gives it `TRANSIENT_ATTACHMENT` usage and prefers lazily-allocated memory. The structural half of that test is shared as `IsAttachmentOnlyTexture`, which OpenGL's existing  `IsRenderbufferSufficient` now builds on.
- `HelloOpenXR` demonstrates MSAA (`--samples=N`, or the `debug.llgl.samples`  system property on Android). It also fixes a `hasMultiView` → `hasMultiview`  typo that meant the example did not compile at all — unnoticed because examples  and OpenXR are both off by default.

Known limitation: depth submission (reprojection) is unavailable while multi-sampling, because multi-sampled depth cannot be resolved into the runtime's single-sampled depth image without `VK_KHR_depth_stencil_resolve`. This is reported at runtime rather than silently submitting an unwritten image. A follow-up PR lifts the restriction.

### DIFF
--- a/include/LLGL/SwapChainFlags.h
+++ b/include/LLGL/SwapChainFlags.h
@@ -66,7 +66,7 @@ struct SwapChainDescriptor
     \remarks The final name of the native hardware resource is implementation defined.
     \see RenderSystemChild::SetDebugName
     */
-    const char*     debugName       = nullptr;
+    const char*     debugName           = nullptr;
 
     /**
     \brief Screen resolution (in pixels).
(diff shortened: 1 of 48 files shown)


sources/Renderer/Vulkan/VKCore.cpp
```diff
@@ -334,6 +334,16 @@ VkFormat VKFindSupportedImageFormat(VkPhysicalDevice device, const VkFormat* can
     LLGL_TRAP("failed to find suitable image format");
 }
 
+bool VKHasMemoryType(const VkPhysicalDeviceMemoryProperties& memoryProperties, std::uint32_t memoryTypeBits, VkMemoryPropertyFlags properties)
```
LukasBanana: This is essentially a copy of `VKFindMemoryType()`. Let's generalize this by moving this into a static function which is used by both `VKFindMemoryType()` and `VKHasMemoryType()`, something that returns `~0u` instead of trapping to signal an invalid output.

include/LLGL/SwapChainFlags.h
```diff
@@ -74,6 +75,17 @@ struct SwapChainDescriptor
...
+    /**
+    \brief Preferred color format for the swap-chain buffers. By default Format::Undefined.
+    \remarks If this is Format::Undefined, the renderer picks a format automatically (the default behavior).
+    Otherwise the renderer uses this format if the presentation surface supports it, and falls back to its
+    automatic selection if it does not. This is primarily intended to request an sRGB format (e.g.
+    Format::BGRA8UNorm_sRGB), so the hardware performs the linear-to-sRGB conversion on write.
+    To determine the actual color format of a swap-chain, use the SwapChain::GetColorFormat function.
+    \see SwapChain::GetColorFormat
+    */
+    Format          format          = Format::Undefined;
```
LukasBanana: If we add a new interface field that makes the existing ones redundant, I would like us to consider refactoring them. The concept of specifying color and depth bits to let the backend choose its optimal format is a remnant from the OpenGL backend, where you could only select those bits in the `PIXELFORMATDESCRIPTOR` struct but all new APIs let you select an actual format. I would also suggest we make this a list (i.e. `DynamicVector` or `SmallVector<Format, 1u>`) and name it `preferredColorFormats` and also provide `preferredDepthStencilFormats` to make it clear they are only preferences but not guaranteed and the actual final formats can be queried as before via `SwapChain::GetColorFormat()`/`GetDepthStencilFormat()`. The depth-stencil format list would need a default entry as substitute for the default depth and stencil bits so that an empty list can be interpreted as a disabled depth-stencil.

If you agree, I'm okay if you only implement this for the Vulkan backend and I'll take care of all the other backends, but let's first discuss what makes sense here when we expand the public interface.

sources/Renderer/Vulkan/VKSwapChain.cpp
```diff
@@ -711,6 +712,25 @@ VkSurfaceFormatKHR VKSwapChain::PickSwapSurfaceFormat(const std::vector<VkSurfac
     if (surfaceFormats.empty())
         LLGL_TRAP("no Vulkan surface formats available");
 
+    /* Honor an explicitly requested format when the surface supports it.
+       This is how an application asks for an sRGB swap-chain, so the hardware performs the
+       linear-to-sRGB conversion on write instead of the shader. */
+    if (requestedColorFormat_ != Format::Undefined)
+    {
+        const VkFormat requestedVkFormat = VKTypes::Map(requestedColorFormat_);
```
LukasBanana: If we don't use this `Format` anywhere else, we could also just store this as `VkFormat` so we have to map it from `LLGL::Format` only once in the constructor. Same if we go the route with a list of preferred formats.

LukasBanana (changes requested):
Let's discuss how we proceed with refactoring the swap-chain's color and depth/stencil bits. This is a good opportunity to sunset the old bit parameters and replace them with lists of preferred formats.

include/LLGL/SwapChainFlags.h
joevdh: Ok, so you're thinking remove colorBits, depthBits etc?  Wasn't sure if you'd be ok with breaking API back-compat.  

What would you think of a more RAII approach where RenderSystem::CreateSwapChain() fails and returns null if the SwapChainDescriptor passed in is targeting formats that the hardware doesn't support?  Then we can avoid adding a DynamicVector to SwapChainDescriptor.

LukasBanana: We'd deprecate `colorBits` etc. and eventually remove them. You can ignore that deprecation part for now, we can do that once our new approach is implemented in all backends.

Maybe I'm overthinking this with using a list of preferences and perhaps it's more straightforward to have a single format for both color and depth-stencil. I would like to keep falling back to a default value or next best format, though, instead of failing with an unsupported format. The user could still check the final format via `GetColorFormat()` etc. and match it with the requested format to decide whether it's considered 'failed'. Moreover, we should also provide a list of supported swap-chain formats analogous to `RenderingCapabilities::textureFormats`, so the client doesn't have to do guess work until a supported swap-chain is successfully created.

So let's add the following:
- List of supported swap-chain formats in `RenderingCapabilities`: `swapChainColorFormats` and `swapChainDepthStencilFormats`.
We leave them empty in the backends that don't implement the new approach yet.
- Requested formats in `SwapChainDescritpor` as you suggested but for both color and depth-stencil: `colorFormat` (defaults to `RGBA8UNorm`) and `depthStencilFormat` (defaults to `D24UNormS8UInt`).

I have to note that depth-stencil formats are not necessarily surface or swap-chain dependent unlike the color formats, since they are created as independent depth-stencil buffer. So maybe we can skip `swapChainDepthStencilFormats` list. What do you think?

joevdh: Sounds good.  Made the changes you suggested and submitted the new revision.  A couple notes:
- I left colorFormat and depthStencilFormat as defaulting to Unknown to indicate that the code show fall back to the previous method for determining the formats.  I figured you could change the default values once you're ready to deprecate the old bits parameters
- Updated the Vulkan path to use the new settings, but didn't go all the way on the other backends

LukasBanana:
Awesome work! I'm glad to see the old bit parameters deprecated soon.

